PR #4, opened 2015-06-13
SINGA-7 Implement shared memory Hogwild algorithm

nudles:
Add support for shared memory Hogwild [1] training in single process (node).
Hogwild training is enabled by setting share_memory to true (i.e., worker groups share memory space for parameter values) in cluster.conf.
Depending on the configurations of cluster.conf, there are two places to do the update:

1. if sever_update is true (default) and there is at least one server group, workers send gradients to servers
which conduct the update based on sending worker's running iteration.
2. otherwise, workers update parameters locally.

The first group conducts the training, validation and test. The training performance is reported periodically for the first group. Validation and test performance is reported immediately. Training performance for other groups are omitted currently.

[1]B. Recht, C. Re, S. J. Wright, and F. Niu. Hogwild: A lock-free approach to parallelizing stochastic gradient descent. In NIPS, pages 693–701, 2011.


nudles:
To run the cifar10 example using the shared memory Hogwild, just configure the cluster.conf as

1. Do update at worker side with 2 worker groups
    nworker_groups: 2
    nserver_groups: 0
    nworkers_per_group: 1
    nworkers_per_procs: 2
    workspace: "examples/cifar10/" 

2. Or do update at the server side with 2 worker groups
    nworker_groups: 2
    nserver_groups: 1
    nservers_per_group:1
    nworkers_per_group: 1
    nworkers_per_procs: 2
    workspace: "examples/cifar10/"

PR #5, opened 2015-06-14
SINGA-6 Update the implementation of singletons to make them thread-safe

nudles:
Thanks to Xuan Liu (from google) for pointing out that the Singleton implementation in SINGA is not thread-safe. Discussions on efficient thread-safe singleton are available at http://stackoverflow.com/questions/2576022/efficient-thread-safe-singleton-in-c. 

This pull request updates the utils/singleton.h file to make the two singletons thread-safe.
In specific, for the Singleton class, we simple use C++11's static construction as suggested in [Stackoverflow](http://stackoverflow.com/a/19907903).
The ASingleton (constructor with Argument) was used for sharing the Mshadow::Random among Layers and Params. To make it thread-safe, 

1. we change it to TSingleton (Thread specfic singlton).
2. we add a construtor without arguments for Mshadow::Random which uses a seed generated based on the system clock.
3. we replace the rand/srand in Mshaodw::Random<cpu> with C++11's random functions, i.e., from random.h

Now each thread will have a separate Mshadow::Random object if it calls TSingleton<Random<cpu>>::Instance() to generate random numbers.
Hence it is thread-safe.


nudles:
To avoid frequent changes of the configuration files, cifar10 example's configuration files are renamed to cluster.conf.example and model.conf.example. Users are suggested to copy *.conf.example to *.conf to run the program. TODO add *.conf in .gitignore to ignore the cluster.conf and example.conf.

PR #10, opened 2015-06-17
SINGA-17 Add root permission within thirdparty/install.

xiezl:


nudles:
Previously, users has to call `sudo ./install.sh <software_name>`  to install the software into system's default lib folder. However, it is not necessary to extract the tar files and call `make` command with sudo right. Only make install requires sudo right. This pull request improves the install.sh script.
Now, if users do not provide the installation prefix, the install.sh script will call 
    
    sudo make install ...

to install the software in system's default lib path. Otherwise it calls

    make install...

to install the software in user specified PREFIX folder.

Another issue with the previous script is that if users call `sudo ./install.sh zookeeper`, then the zookeeper folder cannot be written by the ./bin/singa-run script without sudo right. Consequently, the zookeeper cannot be started successfully at run time.

PR #17, opened 2015-06-23
SINGA-19 Slice large Param objects for load-balance

nudles:
Some Param objects in deep learning models are much larger than other Param objects. For example, a weight matrix is usually 100 times larger than a bias vector. The difference in Param size causes two problems,

1. if there are multiple servers in one server group, then the servers may be assigned different number of parameters to update.
2. if there are multiple server groups, e.g., in distributed Hogwild framework, then these server groups may be assigned different number of parameters to maintain.

This ticket its to slice large Param objects to solve the load-balance problem. 
A simple slicing algorithm is used to make it possible to partition all sliced objects into m boxes such that all boxes are assigned similar size of parameters. Because these sliced objects may be partitioned among servers within one group and among server groups,  m is the least common multiplier of nserver_per_group and nsever_groups.

The slicing algorithm iterates over every box and fills in Param objects if it has enough space. The box capacity is total_param_number/m. But to avoid small slices, it may assign at most m/10 additional space for one box to include a complete Param object. For example, if one box has 100 free space, but the next Param object is 150 and m=1000. In such case, it will put this Param object into the box because 100+1000/10=200>150. Consequently, it avoids slicing a Param object.

The Param class is updated to include some fields for the slices (one Param object may have multiple slices). The code for workers is almost untouched, but the code for the stub thread has been changed a lot to convert a Param object into multiple slices. The Server treats one slice as a Param object, hence there is not much change for its processing logic except that it has to use the slice ID from a message as Param ID to look up its  param shard.

### DIFF
--- a/examples/cifar10/model.conf
+++ b/examples/cifar10/model.conf
@@ -1,8 +1,9 @@
 name: "cifar10-convnet"
 train_steps: 1000
-test_steps:100
+test_steps:10
 test_frequency:300
 display_frequency:30
+alg: kBackPropagation
 updater{
   momentum:0.0
   weight_decay:0.004
--- a/include/communication/msg.h
+++ b/include/communication/msg.h
@@ -39,12 +39,14 @@ class Msg {
   inline void SwapAddr() { std::swap(src_, dst_); }
   inline void set_type(int type) { type_ = type; }
   inline int type() const { return type_; }
-  inline void set_target(int first, int second) {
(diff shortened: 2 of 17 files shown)


nudles:
A simple way to test this feature is to configure the server group with >1 servers in cluster.conf for cifar10 example:

    nservers_per_group: 2
    nservers_per_procs: 2
    workspace: "examples/cifar10/"

PR #18, opened 2015-06-25
SINGA-8 Implement distributed Hogwild

nudles:
Generally, both the Downpour framework of Google Brain [1] and the Caffe's distributed Hogwild implementation are extensions of the shared memory Hogwild training. In this pull request, we refer to the second one.

In specific, each server group masters a subset of parameters. When synchronizing with other server groups, it aggregates all updates for its subset from other server groups and sends back (e.g., broadcast) the updated parameters back to all other server groups. The synchronization is conducted asynchronously. The synchronization is triggered as long as the network bandwidth is not fully utilized.

[1]J. Dean, G. Corrado, R. Monga, K. Chen, M. Devin, Q. V. Le, M. Z. Mao, M. Ranzato, A. W. Senior, P. A. Tucker, K. Yang, and A. Y. Ng. Large scale distributed deep networks. In NIPS, pages 1232{1240, 2012.


nudles:
To run the distributed Hogwild framework, configure the cluster.conf as:

nworker_groups: 2
nserver_groups: 2

and start one process as,
    
    ./bin/singa-run.sh -model=... -cluster=...

and then start another process as,
 
    ./singa -model=... -cluster=...

Note that the two command are different! The first one will start the zookeeper.
TODO, create a script to start multiple processes.

PR #20, opened 2015-06-26
SINGA-9 Add Support for Restricted Boltzman Machine (RBM) model

lzjpaul:
This ticket is to implement RBM in SINGA.
To training RBM models, the Contrastive Divergence (CD) algorithm should be implemented. 
We have implemented a BPWorker to run the Back-Propagation algorithm. To implement the CD algorithm, we follow the same way to create a CDWorker whose RunOneBatch function controls the logic of the CD algorithm, including positive phase, negative phase and computing gradient phase.
RBM's layers are different to the layers for feed-forward neural networks, hence new layers for RBM models should be added.


src/trainer/worker.cc
```diff
@@ -331,4 +331,108 @@ void BPWorker::TestOneBatch(int step, Phase phase, shared_ptr<NeuralNet> net, Me
...
+      for (shared_ptr<Param> p : layer->GetParams()) {
+        Update(p, step);
+      }
+  }
+}
+
+void CDWorker::LossPhase(int step, Phase phase,
+     shared_ptr<NeuralNet> net, Metric* perf) {
+  auto& layers = net->layers();
+  if (phase == kTrain) {
```
nudles: can you merge the train phase and test phase since their code are very similar except some condition checks?

src/trainer/worker.cc
```diff
@@ -331,4 +331,108 @@ void BPWorker::TestOneBatch(int step, Phase phase, shared_ptr<NeuralNet> net, Me
...
+  if (step % modelproto_.vis_step() == 0 && step!= 0) { /*print weight matrix*/
+    BlobProto bp;
+    int rownum;
+    int colnum;
+    for (auto& layer : layers) {
+      if (layer->is_bottomlayer()) {
+        for (shared_ptr<Param> p : layer->GetParams()) {
+          rownum = p->data().shape()[0];
+          colnum = p->data().shape()[1];
+          Tensor<cpu, 2> weight(p->mutable_cpu_data(), Shape2(rownum, colnum));
```
nudles: No need to create an tensor object and do a nested loop to fill bp.
Single loop for(int i=0;i<p->size();i++) is enough.

src/trainer/worker.cc
```diff
@@ -331,4 +331,108 @@ void BPWorker::TestOneBatch(int step, Phase phase, shared_ptr<NeuralNet> net, Me
     }
 }
 
+/****************************CDWorker**********************************/
+CDWorker::CDWorker(int thread_id, int group_id, int worker_id):
+  Worker(thread_id, group_id, worker_id) {
+}
+
```
nudles: add some TODO comment to say that model partition (spliting layers) will be done in the future.

src/trainer/worker.cc
```diff
@@ -331,4 +331,108 @@ void BPWorker::TestOneBatch(int step, Phase phase, shared_ptr<NeuralNet> net, Me
...
+      // clock_t s=clock();
+    layer->ComputeFeature(kPositive);
+  }
+}
+
+void CDWorker::NegativePhase(int step, shared_ptr<NeuralNet> net) {
+  auto& layers = net->layers();
+  for (int i = 0; i < modelproto_.pcd_k(); i++) {
+    for (auto& layer : layers) {
+      if (layer->is_bottomlayer() || layer->is_toplayer())
```
nudles: add some comment for the check ?

src/proto/model.proto
```diff
@@ -285,6 +295,18 @@ message DropoutProto {
...
 
+// Message that stores parameters used by DBMBottomLayer
+message DBMBottomProto {
+  optional uint32 num_output = 1; // The number of outputs for the layer
+  optional bool bias_term = 2 [default = true]; // whether to have bias terms
+}
+
+// Message that stores parameters used by DBMTopLayer
+message DBMTopProto {
+  optional uint32 num_output = 1; // The number of outputs for the layer
```
nudles: uint32 - >int32 ? since other params use int32.

include/neuralnet/layer.h
```diff
@@ -85,6 +85,112 @@ class DropoutLayer: public Layer {
...
+    CHECK_LT(k, srclayers_.size());
+    return kOneToAll;
+  }
+
+  virtual void ComputeFeature(Phase phase,
+     const vector<shared_ptr<Layer>>& srclayers);
+  virtual void ComputeGradient(const vector<shared_ptr<Layer>>& srclayers);
+  virtual void ComputeLoss(Metric* perf);
+  virtual Blob<float>* mutable_data(const Layer* from) {
+    if (kPhase)
```
nudles: where is kPhase defined?
variables starts with 'k' is usually used as constants.
you may want to change mutable_data as 
... mutable_data(const Layer* from, Phase phase=kTrain)
and then check using phase. But all mutable_data function should be changed in layer.h and base_layer.h.

include/neuralnet/layer.h
```diff
@@ -85,6 +85,112 @@ class DropoutLayer: public Layer {
...
+    return vector<shared_ptr<Param>>{bias_};
+  }
+
+ private:
+  //! dimension of the visible layer
+  int vdim_;
+  int batchsize_;
+  // batchsize of negative phase
+  int neg_batchsize_;
+  bool is_first_iteration_top;
```
nudles: use xxx_ for member variable.

src/neuralnet/layer.cc
```diff
@@ -160,6 +160,200 @@ void DropoutLayer::ComputeGradient(const vector<SLayer>& srclayers)  {
...
+    kPhase = false;
+    CHECK_EQ(srclayers[0]->data(this).count(), neg_batchsize_*vdim_);
+    Tensor<cpu, 2> negsrc(srclayers[0]->mutable_data(this)->mutable_cpu_data(),
+         Shape2(neg_batchsize_, vdim_));
+    Tensor<cpu, 1> bias(bias_->mutable_cpu_data(), Shape1(vdim_));
+    negsrc+=repmat(bias, neg_batchsize_);
+    negsrc = F<op::sigmoid>(negsrc);
+    for (int i = 0; i < neg_batchsize_; i++)
+      for (int j = 0; j < vdim_; j++)
+        negsrc[i][j] = static_cast<float>((rand() / static_cast<double>(RAND_MAX)) > negsrc[i][j] ? 0 : 1);
```
nudles: replace manual sampling with mshadow's sample function.

src/neuralnet/layer.cc
```diff
@@ -160,6 +160,200 @@ void DropoutLayer::ComputeGradient(const vector<SLayer>& srclayers)  {
...
+            Shape2(neg_batchsize_, hdim_));
+          Tensor<cpu, 2> weight(weight_->mutable_cpu_data(),
+            Shape2(vdim_, hdim_));
+          kPhase = true;
+          CHECK_EQ(srclayers[0]->data(this).count(), batchsize_*vdim_);
+          Tensor<cpu, 2> possrc(srclayers[0]->mutable_data(this)->mutable_cpu_data(),
+          Shape2(batchsize_, vdim_));
+          Tensor<cpu, 2> negsrc(negsrc_.mutable_cpu_data(),
+            Shape2(neg_batchsize_, vdim_));
+          for (int i = 0; i < batchsize_; i++)
```
nudles: use mshadow's copy function.

include/neuralnet/layer.h
```diff
@@ -85,6 +85,112 @@ class DropoutLayer: public Layer {
...
+  //! dimension of the visible layer
+  int vdim_;
+  int batchsize_;
+  // batchsize of negative phase
+  int neg_batchsize_;
+  bool is_first_iteration_bottom;
+  float scale_;
+  shared_ptr<Param> weight_, bias_;
+  // data to store variables of negative phase
+  Blob<float> hidden_data_;
```
nudles: better rename hidden_data_ to neg_hid_ or neg_data_, possrc to pos_vis, negsrc to neg_vis
just to make it consistent with the terminologies used in papers, or follow those used in the http://deeplearning.net/tutorial/rbm.html

src/neuralnet/layer.cc
```diff
@@ -160,6 +160,200 @@ void DropoutLayer::ComputeGradient(const vector<SLayer>& srclayers)  {
...
+  weight_->Setup(proto.param(0), vector<int>{vdim_, hdim_}, vdim_*hdim_);
+  bias_->Setup(proto.param(1), vector<int>{vdim_}, 0);
+  srand((unsigned)time(NULL));
+}
+void DBMBottomLayer::SetupAfterPartition(const LayerProto& proto,
+      const vector<int> &shape,
+      const vector<SLayer>& srclayers) {
+}
+
+void DBMBottomLayer::ComputeFeature(Phase phase, const vector<SLayer>& srclayers) {
```
nudles: can we compress the code by first define some pointers for neg phase or pos phase, and them use these pointers to create tensors and do the dot, sigmoid operations.

src/neuralnet/layer.cc
```diff
@@ -160,6 +160,200 @@ void DropoutLayer::ComputeGradient(const vector<SLayer>& srclayers)  {
...
+  Tensor<cpu, 2> weight(weight_->mutable_cpu_data(), Shape2(vdim_, hdim_));
+  Tensor<cpu, 1> bias(bias_->mutable_cpu_data(), Shape1(vdim_));
+  Tensor<cpu, 2> reconstruct(Shape2(batchsize_, vdim_)); /*reconstruct error*/
+  AllocSpace(reconstruct);
+  reconstruct = dot(data, weight.T());
+  reconstruct+=repmat(bias, batchsize_);
+  reconstruct = F<op::sigmoid>(reconstruct);
+  for (int i = 0; i < batchsize_; i++)
+    for (int j = 0; j < vdim_; j++) {
+      loss += -(possrc[i][j]*log(reconstruct[i][j])+(1-possrc[i][j])*log(1-reconstruct[i][j]));
```
nudles: replace to single loop and avoid using mshaow's [] operator. try to use mshadow's tensors.

src/trainer/worker.cc
```diff
@@ -331,4 +331,108 @@ void BPWorker::TestOneBatch(int step, Phase phase, shared_ptr<NeuralNet> net, Me
...
+        bp.set_width(colnum);
+        auto cluster = Cluster::Get();
+        char *n_str = new char[70];
+        // const char *str = cluster->_folder();
+        std::string str = cluster->vis_folder() + "/";
+        std::string s = std::to_string(step);
+        char* buffer = (char*) s.c_str();
+        strcpy(n_str, (char*)str.c_str());
+        strcat(n_str, buffer);
+        WriteProtoToBinaryFile(bp, n_str);
```
nudles: compress above lines to 
string filename=cluster->vis_folder()+"/"+std::to_string(step); // may also add param name in the filename
WritePRotoToBinary(bp, filename.c_str());

PR #27, opened 2015-07-10
Singa-GPU-1 add xpu in blob

DotComing:
Singa-GPU-1 add xpu in blob from Netease.


DotComing:
Pull request cancelled.

PR #28, opened 2015-07-11
SINGA-29 Update NeuralNet class to enable layer partition type customization

nudles:
1. Clean the code for NeuralNet and Graph classes.
   Graph class only provides functions about Node and Edge management, e.g., add, remove and toplogy sort.
   NeuralNet provides one function (CreateGraph) to convert net configuration into a Graph. Net partitioning
   is done in CreateGraph function. The CreateNetFromGraph function create and connect layers from the graph.
2. Users can customize the partition for whole net and for a specific layer through field partition_dim of LayerProto and NetProto.
   the configuration of LayerProto overwrites that of NetProto.

Tested on single process with non-distributed training, shared-memory hogwild, one worker group with 2 workers.
Tested with two processes for downpour and distributed hogwild. Downpour has similar performance as shared-memory hoglwild; while Distributed hogwild does not perform as good non-distributed training.

### DIFF
--- a/include/neuralnet/base_layer.h
+++ b/include/neuralnet/base_layer.h
@@ -1,5 +1,5 @@
-#ifndef SINGA_BASE_LAYER_H_
-#define SINGA_BASE_LAYER_H_
+#ifndef SINGA_NEURALNET_BASE_LAYER_H_
+#define SINGA_NEURALNET_BASE_LAYER_H_
 
 #include <vector>
 #include <string>
@@ -14,7 +14,7 @@
 #include "utils/common.h"
 #include "utils/blob.h"
 
-namespace singa{
+namespace singa {
 
 using std::vector;
 using std::string;
@@ -31,8 +31,8 @@ class Layer;
  */
 class Layer {
  public:
-  Layer(){}
-  virtual ~Layer(){}
+  Layer() { }
+  virtual ~Layer() {}
   /**
    * Setup layer properties.
    *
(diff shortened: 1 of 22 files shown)


nudles:
Many shared pointers are replaced with raw pointers to make the code tidy. 'shared_ptr' adds many characters that usually leads to two lines for one statement.

This pull request also cleans the API of Layer class:
* Setup(LayerProto, int npartitions), setup the layer's member data according to user specified configuration. If npartitions > 1, then this layer should be setup as one of npartitions sub-layers.

* ComputeFeature(Phase, Metric*), now the this function can be used for forward propagation of BP algorithm by passing Phase as kTrain or kForward in BPWorker; It can also be used for the positive phase of CD algorithm by passing Phase as kPositive. The Metric pointer is to collect performance information from this layer. E.g., the loss layers would add metric like loss or accuracy to it.

* ComputeGradient(Phase), the Phase argument is not used now (neither BP nor CD need this argument), but is added just in case some other algorithms may need it in the future.

* DebugString(int, Phase), this function is newly added to generate some debug information from one layer. The DebugInfo function of NeuralNet class is removed.

* ConnectionType dst_layer_connection(), return the connection type of this layer and all dst layers.
  Currently support two connections: kOneToOne and kOneToMany. kOneToOne indicates the users implement the ComputeFeature and ComputeGradient function considering only one dest layer. In this case, a SplitLayer will be added automatically to connect this layer with all dest layer. kOneToMany indicates the users has already considered multiple dest layers in the implementation. Hence no SplitLayer will be inserted.

* ConnectionType src_neuron_connection(int k)， return the connection type between one neuron of this layer and its source layer. Currently support two connection types: kOneToOne, and kOneToAll.
kOneToOne indicates the neuron depends on only one neuron from src layer. kOneToAll indicates the neuron depends on all neurons from src layer. param k is the index of source layer (current only support k = 0).

The Metric Class is updated to use Google Protocol buffer as the serialization tool.
* ToLogString() will print a human readable string, which would be used for logging.
* ToString() will serialize the metric object into binary string for transmission.

Now the metrics is collected from every layer's ComputeFeature function. Hence the Worker does not need to explicitly collecting metrics from loss layers.

PR #33, opened 2015-07-18
SINGA-12 Supprt Checkpoint and Restore

nudles:
The checkpoint is done in the Worker class and controlled by two model configuration fields: checkpoint_after and checkpoint_frequency.

Only do checkpoint for the Params owning the param values from the first group. The name, version and values of one Param are dumped onto disk (the path is <workspace>/checkpoint/step<training step>-worker<worker id>.bin). It is possible that the snapshot is separated into multiple files because the neural net is partitioned into multi workers.

The checkpoint files can be used:

  * application 1: to restore (resume) the training by setting the command line argument -resume = true.
The Resume function of the Trainer will find the files for the latest snapshot and add them to the model.conf's checkpoint filed. It also set the model config's step field to the snapshot step (extracted from file name).

  * application 2: as the pre-training result of another model. Users have to config the new model's checkpoint field to add the paths of all checkpoint files of the same snapshot.

The Worker's InitLocalParam will init Param's from checkpoint files if available. Otherwise it randomly init them using user configured init method. It matches the Param objects based on name. If the Param is not configured with a name, NeuralNet class will automatically create one for it based on the name
of the layer to which the Param belongs. In this case,

  * for application 2, users have to configure the names of the new model params carefully to
match the names of params from the pre-trained model

  * for application 1, the worker-server topology cannot be changed.

Restore for params which are partitioned due to model partitioning is not supported. Because if the pre-training is done using 2 workers, while the new model is trained with 3 workers,
then the same original param is partitioned in different ways and hence cannot be matched.

TODO: update the sing-run.sh to include the command line argument for resume, as 

    ./bin/singa-run.sh -conf=.. . -resume=...

### DIFF
--- a/bin/singa-run.sh
+++ b/bin/singa-run.sh
@@ -25,7 +25,7 @@
 
 usage="Usage: \n \
   (single node): singa-run.sh -cluster=YOUR_CONF_FILE -model=YOUR_CONF_FILE \n \
-  (distributed): singa-run.sh -conf=YOUR_CONF_DIR \ 
+  (distributed): singa-run.sh -conf=YOUR_CONF_DIR \
   (the directory should contain cluster.conf/model.conf/hostfile)"
 
 #if [ $# -le 0 ] || [ $# -ge 3 ] ; then
@@ -51,7 +51,7 @@ fi
 
 if [ $valid_args = false ] ; then
   echo -e $usage
-  exit 1 
+  exit 1
 fi
 
 # get singa-base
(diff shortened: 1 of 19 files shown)


nudles:
For the  cifar10 example, if we configure the model.conf as:

    ...
    checkpoint_after: 100
    checkpoint_frequency: 300
    ...

Then after training for 700 steps, under cifar10/checkpoint folder, there would be two checkpoint files:

    step400-worker0.bin
    step700-worker0.bin

We can resume the training from the last checkpoint (i.e., step 700) by:
 
    ./bin/singa-run.sh -conf=examples/cifar10 -resume=true

We can also use the the snapshot from step 400 as the pre-trained model for a new model by configuring the model.conf of the new model as:

    ...
    checkpoint : examples/cifar10/checkpoint/step400-worker0.bin
    ...

the launch command is the same as starting a new job
  
    ./bin/singa-run.sh -conf=...

PR #56, opened 2015-08-18
SINGA-61 Support user defined classes

nudles:
To make SINGA extensible, polymorphism is used for many classes, e.g., Layer, Updater. Users configure a type field to decide which subclass to use at run time.
E.g., the LayerProto has a type field. But this type field can only be set to built-in layer types.
This commit adds a user_type field for configuring user defined layers. E.g.,

    LayerProto {
      optional string user_type = 10;
    }

A helper function Layer::Create(const LayerProto&) is added to create a Layer object based on user configuration.
Similar updates are applied for other classes, namely, Param, Worker, Updater.

Driver class in declared in driver.h.
singa.h now has only include directions.


nudles:
Please re-create data shards for mnist and cifar10 examples, as the protocol buffer message for Record has been changed.

PR #59, opened 2015-08-22
SINGA-21 Code review 4

wangsheng1001:
review param.h, param.cc
  - ShareFrom(): init vectors by resizing instead of assignning
  - add a summary for put/get/update/sync request workflows
  - implement reserve flags correctly for all HandleXX functions
    TODO: remove all reserve-flag checks later
  - format the code

format update.h, updater.cc

review base_layer.h/cc, layer.h/cc, optional_layer.h/cc
  - change basic functions in Layer from virtual to inline
  - make ComputeFeature and ComputeGradient have same paramters
    i.e. ComputeGradient(Phase phase, Metric* perf)
  - add a middle level of layer category for ConnectionLayer and NeuronLayer
  - re-arrange Layers in these file
    base_layer.h contains layer categories that are extended by other layers
    layer.h contains layer implementations


nudles:
need to update the Makefile.am to compile new files.

PR #62, opened 2015-09-06
SINGA-67 Add singatest into build targets

xiezl:
Modified Makefile.am to add a target named "singatest".
Now "make" will generate 5 files/libs: libsinga.la libgtest.la singa singatool singatest.
User can also specify the target he wants to generate, for example, "make singa".

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,9 +4,12 @@ AUTOMAKE_OPTIONS = foreign subdir-objects
 AM_CPPFLAGS = -I$(top_srcdir)/src
 
 MSHADOW_FLAGS = -DMSHADOW_USE_CUDA=0 -DMSHADOW_USE_CBLAS=1 -DMSHADOW_USE_MKL=0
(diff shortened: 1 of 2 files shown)


nudles:
could you separate the test part with the core part?

particularly, we provided three compiling commands,

    make

compiles the core part.

    make test

compiles the test part.

    make all

compiles both.

Would it be better?
Thank you.

xiezl:
updated with your comments and SINGA-69.

nudles:
the compilation generates many screen outputs...
it would be better to hide these screen outputs like the previous version.

rename `enable-singatest` to `enable-test`?

xiezl:
All updated.

nudles:
summary:

1.  compile with `-g`
     
       ./configure --enable-debug
        make

2. compile both singa core and test code
  
        ./configure --enable-test
        make

    or

        ./configure 
        make  // compile core only
        make singatest   // compile test only

PR #67, opened 2015-09-15
SINGA-57 Improve Distributed Hogwild

nudles:
The Trainer::Run and Server::Run was messy due to the monitoring of the network bandwidth for implementing the distributed hogwild. This pull request re-implement this framework removing the bandwidth monitoring.

The ClusterProto::sync_freq field controls the frequency of sync between
server groups.
After updating of Param (slice), the server checks the num of updates
since last sync. It also checks the num of pending syncs (i.e., requests
haven't received reponses) to avoid sending too many msgs to stopped
servers (the msgs would be occupy the memory of the sending buffer)
The server respones to every sync requests with the latest Param values.

Note: current does not support (there is bug) multiple worker groups in
one process for the distributed hogwild framework. We recommend to
replace this cluster topology with in-memory hogwild, i.e., launching
one worker group with multiple workers and one server group.

### DIFF
--- a/examples/cifar10/job.conf
+++ b/examples/cifar10/job.conf
@@ -2,7 +2,7 @@ name: "cifar10-convnet"
 train_steps: 1000
 test_steps: 100
 test_freq:300
-disp_freq:30
+disp_freq: 30
 train_one_batch {
   alg: kBP
 }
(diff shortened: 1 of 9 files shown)


nudles:
an example cluster topology of distributed hogwild is 

    nworker_groups: 2
    nserver_groups: 2
    
Each server group serves one worker group. One worker group and one server group will be created in one processes (two processes will be launched in total).

PR #68, opened 2015-09-16
SINGA-10 Add Support for Recurrent Neural Networks (RNN)

nudles:
Implement RNN for language modelling following http://www.fit.vutbr.cz/~imikolov/rnnlm/.

Need to test the distributed training frameworks on example models to ensure the singatool works correctly.


nudles:
Follow the instruction in examples/rnnlm/README.md to do test.

PR #73, opened 2015-09-30
SINGA-77 Integrate with Apache RAT

ug93tad:
(Thanks Zhongle for the help with the Makefile.am)

Apache RAT enables automatic checking of license headers in source files. SINGA supports RAT by adding
a target into the Makefile, and providing a `rat-excludes` file at the top-level directory.
Specifically, the `make rat` target does the following:
+ Check for java and Apache RAT .jar file in the $RAT_PATH variable.
+ Invoke `make clean`
+ Invoke RAT and write the output to `rat_check` file.

Notice the side-effect of `make rat` that removes all built files. Thus, it is recommended to use before
building the source.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -163,7 +163,20 @@ endif
 clean-local:
 	rm -rf $(PROTO_SRCS) $(PROTO_HDRS)
 	rm -rf $(PROTO_PYS)
+	rm -rf rat_check
+	rm -rf tool/pb2
 
+rat:
+	@if test ! -z '$(shell command -v java 2>/dev/null)'; then \
+		if test ! -z '$(shell echo $$RAT_PATH)'; then \
+			make clean;\
+			java -jar $(RAT_PATH) -E rat-excludes -d . > rat_check; \
+		else \
+			echo "RAT_PATH is not set to correct jar file"; \
+		fi \
+	else \
+		echo "java is not found"; \
+	fi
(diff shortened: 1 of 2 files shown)


nudles:
Would it be better to print the output on the screen?
Did you see any other temporary or binary files left by `make clean` except .dirstamp?

rat-excludes
```diff
@@ -0,0 +1,15 @@
+rat-excludes
+Makefile.*
+configure
+.gitignore
+conf/*
+config/*
+examples/*
+include/utils/tinydir.h
+tool/cpplint.py
```
nudles: Should we exclude tinydir.h, cpplint.py and examples/?

Makefile.am
```diff
@@ -163,7 +163,20 @@ endif
...
+	rm -rf rat_check
+	rm -rf tool/pb2
+
+rat:
+	@if test ! -z '$(shell command -v java 2>/dev/null)'; then \
+		if test ! -z '$(shell echo $$RAT_PATH)'; then \
+			make clean;\
+			java -jar $(RAT_PATH) -E rat-excludes -d . > rat_check; \
+		else \
+			echo "RAT_PATH is not set to correct jar file"; \
```
nudles: Maybe we can add a url of the apche rat page for users to download the .jar file.

PR #84, opened 2015-10-18
SINGA-89  Add Docker support

ug93tad:
Docker is a Linux container which is fast to deploy and incurs small overhead. A Docker container functions like a light-weight virtual machine and runs in its own isolated environment. There are at least two benefits in building adding Docker support for SINGA:

+ Out of the box deployment of SINGA: the image, once built, contains the complete environment necessary to start SINGA.

+ Light-weight development and testing environment for distributed features. The image can be used to set up a test-bed consisting of many independent nodes (each node has its own IP address), without the need for cluster hardware.

We add to Github a Dockerfile specifying how to build SINGA image, from which the user can construct the image by executing:

sudo docker build -t singa/base .

The user can choose another name beside singa/base . The build process can take a long time, but it needs to be done only at one host and the result image can then be copied to other hosts.

We also add another Dockerfile which adds Mesos and Hadoop ontop of SINGA. This is closely related to SINGA-11. The image created from this Dockerfile is used to set up the distributed test bed.

See the README.md for more details of the images and how to use them.

IMPORTANT We assume that every host has Docker running. Two nodes on the epiC cluster (`ciidaa-c18` and `ciidaa-c19`) are set up with Docker and the pre-built images.

IMPORTANT For now, the Dockerfile for building Mesos+Hadoop+SINGA will pull the latest from ug93tad's (Anh's) SINGA-11 branch. This pull step is removed after SINGA-11 is merged to the project's master branch.

### DIFF
--- a/tool/docker/singa/.bashrc
+++ b/tool/docker/singa/.bashrc
@@ -1,0 +1,8 @@
+export LIBRARY_PATH=/opt/OpenBLAS/lib:$LIBRARY_PATH
+export LD_LIBRARY_PATH=/opt/OpenBLAS/lib:/usr/local/lib:$LD_LIBRARY_PATH
+export CPLUS_INCLUDE_PATH=/opt/OpenBLAS/include:$CPLUS_INCLUDE_PATH
+export PATH=$PATH:/opt/jdk1.8.0_60/bin:/opt/bin
+export JAVA_HOME=/opt/jdk1.8.0_60
+export SINGA_HOME=/root/incubator-singa
+export PATH=$PATH:$SINGA_HOME/bin
+alias ls="ls --color=always"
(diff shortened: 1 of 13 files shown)


tool/docker/singa/.bashrc
```diff
@@ -0,0 +1,8 @@
+export LIBRARY_PATH=/opt/OpenBLAS/lib:$LIBRARY_PATH
```
nudles: will the .bashrc be the same as that under ~/, i.e., it will be executed by shell automatically once we login into a docker node?
this file is similar to that under tool/docker/mesos.
If docker/mesos is installed after docker/singa, then we can remove that file.

nudles:
merged.
please add a documentation page at http://singa.apache.org/docs/index.html.
thanks.

nudles:
merged.
please update the documentation.
we may need to add a note at the beginning to tell users that Mesos runs inside Docker.

PR #85, opened 2015-10-18
SINGA-11 Start SINGA on Apache Mesos

ug93tad:
Apache Mesos is a fine-grained cluster management framework which enables resource sharing in the same cluster. Mesos abstracts out the physical configurations of cluster nodes, and presents resources to the users in the form of "offers". SINGA uses Mesos for two purposes:

1. To acquire necessary resources for training the model.
2. To launch and monitor progress of the training task.

To this end, we implement a "SINGA Scheduler" which interacts with Mesos master. The scheduler assumes that SINGA has been installed  at the Mesos slave nodes. The scheduler is called when the user wants to start a new SINGA job, and it performs the following steps:

Step 1. Read the job configuration file to determine necessary resources in terms of CPUs, memory and storage.
Step 2. Wait for resource offers from the Mesos master.
Step 3. Determine if the offers meet the requirement of resources.
Step 4. Prepare the task to launch at each slave:
			+ Deliver the job configuration file to the slave node.
			+ Specify the command to run on the	slave: "singa -conf ./job.conf"
Step 5: Launch and monitor the progress

For step 3, we currently implement a simple scheme: the number of CPUs offered by each Mesos slave exceed the total number of SINGA worker and SINGA server per process. In other words, each Mesos slave must be able to run the entire worker group or server group.

For step 4, we currently relies on HDFS to deliver the configuration file to each slave. Particularly, we write the file to a known directory (different for each job) on HDFS and ask the slave to use its Fetcher utility to download the file before executing the task.

The source code and README.md files are in the `tool/mesos` directory.

### DIFF
--- a/tool/mesos/Makefile
+++ b/tool/mesos/Makefile
@@ -1,4 +1,4 @@
-CXX_FLAGS=-I ../include -std=c++11 -I /usr/local/include/hdfs -I /root/incubator-singa/include
+CXX_FLAGS=-I ../include -std=c++11 -I /usr/local/include/hdfs -I ../../include
 LD_FLAGS=-lmesos -lsinga -lhdfs3
 EXE=scheduler
 OBJS=singa_scheduler.o scheduler.pb.o
--- a/tool/mesos/Makefile
+++ b/tool/mesos/Makefile
@@ -1,0 +1,25 @@
+CXX_FLAGS=-I ../include -std=c++11 -I /usr/local/include/hdfs -I /root/incubator-singa/include
+LD_FLAGS=-lmesos -lsinga -lhdfs3
+EXE=scheduler
+OBJS=singa_scheduler.o scheduler.pb.o
+PROTOS=scheduler.proto
+PROTO_HDRS=scheduler.pb.h
+PROTO_SRCS=scheduler.pb.cc
+
+CXX=g++
+
+all: $(PROTO_HDRS) $(EXE)
+
+$(PROTO_SRCS) $(PROTO_HDRS): $(PROTOS)
+	protoc --cpp_out=. $(PROTOS)
+
+$(EXE): $(OBJS)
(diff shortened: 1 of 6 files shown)


tool/mesos/Makefile
```diff
@@ -0,0 +1,25 @@
+CXX_FLAGS=-I ../include -std=c++11 -I /usr/local/include/hdfs -I /root/incubator-singa/include
```
nudles: since mesos is placed inside incubator-singa,
we can replace `/root/incubator-singa` with `../../include`.

tool/mesos/singa.conf
```diff
@@ -0,0 +1,7 @@
+# point to your active zookeeper service
```
nudles: this file is duplicated from `incubator-singa/conf/singa.conf`.
we can use that file and add some comments for configuring the zookeeper host using docker node.

tool/mesos/singa_scheduler.cc
```diff
@@ -0,0 +1,439 @@
...
+ *	Our scheduler performs simply task assignment which guarantees that each process runs an entire work group,
+ *	and each takes all the memory offered by the slave. We assume that each slave runs ONE process, that is the following
+ *	condition holds:
+ *		nCPUs_per_host >= nWorkersPerProcess+nServersPerProcess //if seperate
+										 >= max(nWorkerPerProcess, nServersPerProcess) // if attached
+ */
+using std::string;
+using mesos::SchedulerDriver;
+using std::vector;
+using std::map;
```
nudles: please check the format using
    
    python tool/cpplint.py tool/mesos/*.cc
    python tool/cpplint.py tool/mesos/*.h

nudles:
hadoop-daemon.sh reports 

    chown: missing operand after '/opt/hadoop-2.6.0/logs

Looks like the message is a warning.

PR #87, opened 2015-10-19
SINGA-94 Move call to google::InitGoogleLogging() from Driver::Init() to main().

yeefan:
It appears that google::InitGoogleLogging() was never intended to be called inside a library.
The reason is because calling google::InitGoogleLogging() twice results in program termination with the error log "You called InitGoogleLogging() twice!".
Calling google::InitGoogleLogging() inside Driver::Init() prevents users of the Singa library from making this call in their own main() function, as recommended by glog documentation.

### DIFF
--- a/src/driver.cc
+++ b/src/driver.cc
@@ -64,17 +64,16 @@ extern "C" void openblas_set_num_threads(int num);
 namespace singa {
 
 void Driver::Init(int argc, char **argv) {
-  google::InitGoogleLogging(argv[0]);
-  //  unique job ID generated from singa-run.sh, passed in as "-singa_job <id>"
+  // unique job ID generated from singa-run.sh, passed in as "-singa_job <id>"
   int arg_pos = ArgPos(argc, argv, "-singa_job");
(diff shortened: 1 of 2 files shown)


nudles:
Currently SINGA uses glog for logging.
Hence InitGoogleLogging must be called.
We put it into driver::Init(), therefore users do not need to call google::Init.. explicitly.

I think if SINGA does not depend on glog, then we can remove it from driver::init. Users call google::Init .. if they use glog in their code. 

Doe it make sense?

yeefan:
The thing is, if multiple libraries use glog and each library calls google::InitGoogleLogging, then it becomes impossible to use these libraries together. Personally, I don't like the way glog forces program termination when google::InitGoogleLogging is called multiple. However, without a public API allowing the user to query whether glog has been initialized, the next best course is for all the library writers to cooperate and not call google::InitGoogleLogging inside the library.
By the way, glog does function without a call to google::InitGoogleLogging. In this case, a warning message is printed during the first use of glog without initialization, and all logs just go to stderr without colour.

yeefan:
Thanks again for accepting the pull request. :)

PR #89, opened 2015-10-20
SINGA-95 Add make test after building

xiezl:
Add a new target named "test". Users just run "make test" and the system will finish the compilation and execution of the test.


kaiping:
Hi,

I try to use the new "make test" functionality and it reports an error.

My operations are as follows:
git fetch origin pull/89/head:SINGA-95
git checkout master
git pull origin master
git merge SINGA-95
./autogen.sh
./configure
make
make test

When "make test", an error is reported: 
 CXXLD    singatest
/usr/bin/ld: cannot find -lgtest

Thanks.

xiezl:
Hi kaiping,

The bug is fixed. Pls check.

zl

PR #91, opened 2015-11-05
[SINGA-101] Add ll (ls -l) command in .bashrc file when using docker

YanTangZhai:
Root loads docker containers while the command ll (ls -l) cannot be found. However ll is handier than ls -l.


nudles:
Hi, would you please re-send the pull request with a clean commit based on the latest master branch.
It seems that you merged one old branch with the latest master and then worked on it.

It would be better to follow this work flow

1. checkout the latest master branch to a new branch
2. do the update on the new branch, and commit
3. checkout to the master branch and pull the master from incubator-singa github
4. checkout to the new branch, rebase to the latest master branch 
5. send pull request

PR #93, opened 2015-11-10
SINGA-80 New Blob Level and Address Level Math Operation Interface

jinyangturbo:


ijingo:
In function MVAddCol(383 line in math_blob.h), I think " if (B.transpose()) " should be " if (B->transpose())" because B is a pointer type of Blob.
Maybe the same issue with MVAddRow(423 line).

ijingo:
In some situation, operation that can reduce *column* of A to an element of B is needed like the operation in `Reduce2D`. Maybe explicitly transpose operation can help.

PR #95, opened 2015-11-26
SINGA-104 Add Context Class

seaokcs:
Add New Features:
1 Support for the operation to GPU device
2 Support for the management  of GPU and CPU random number  generator


nudles:
To test this pull request,
1. Copy Makefile.gpu to Makefile
2. Set the cudar_dir and home_dir filed in Makefile to point to your dependency lib folder and cuda folder
3. make singa
4. make test
5. .libs/test

PR #96, opened 2015-11-29
SINGA-97 Add HDFS Store

ug93tad:
This ticket implements HDFS Store for reading data from HDFS. It complements
the existing CSV Store which reads data from CSV file. HDFS is the popular
distributed file system with high (sequential) I/O throughputs, thus supporting
it is necessary in order for SINGA to scale.

HDFS usage in SINGA is different to that in standard MapReduce applications.
Specifically, each SINGA worker may train on sequences of records which do not
lie within block boundary, whereas in MapReduce  each Mapper process a number
of complete blocks.  In MapReduce, the runtime engine may fetch and cache the
entire block over the network, knowing that the block will be processed
entirely. In SINGA, such pre-fetching and caching strategy will be sub-optimal
because it wastes I/O and network bandwidth on data records which are not used.

We defer I/O optimization to a future ticket.

For implementation, we choose `libhdfs3` from Pivotal for HDFS implementation
in C++. This library is built natively for C++, hence it is more optimized and
easier to deploy than the original  `libhdfs` library that is shipped with
Hadoop. libhdfs3 makes extensive use of short-circuit reads to improve local
reads, and it often complain when such option is not set.

Finally, we test the implementation in a distributed environment set up from a
number of  Docker containers. We test with both CIFAR and MNIST examples.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -149,8 +149,6 @@ SINGA_HDRS := include/singa.h \
               include/singa/io/kvfile_store.h \
               include/singa/io/textfile_store.h \
               include/mshadow/cxxnet_op.h \
-              include/singa/io/hdfsfile.h \
-              include/singa/io/hdfsfile_store.h \
               include/mshadow/tensor_expr.h \
               include/mshadow/tensor_container.h \
               include/mshadow/tensor_expr_ext.h \
--- a/Makefile.am
+++ b/Makefile.am
@@ -33,7 +33,7 @@ CFLAGS += $(DEBUG)
 CXXFLAGS += $(DEBUG)
 AC_CXXFLAGS = $(DEBUG)
 
-INCLUDES = -I$(top_srcdir)/include -I/usr/local/include/hdfs
+INCLUDES = -I$(top_srcdir)/include -I/usr/local/include
 
 PROTOS := $(top_srcdir)/src/proto/singa.proto \
           $(top_srcdir)/src/proto/job.proto \
--- a/Makefile.am
+++ b/Makefile.am
@@ -29,11 +29,11 @@ MSHADOW_FLAGS = -DMSHADOW_USE_CUDA=0 -DMSHADOW_USE_CBLAS=1 -DMSHADOW_USE_MKL=0
 DEFAULT_FLAGS = -Wall -pthread -fPIC -std=c++11 -Wno-unknown-pragmas \
               $(MSHADOW_FLAGS) -funroll-loops -DTHREADED
 
-CFLAGS += $(DEBUG)
-CXXFLAGS += $(DEBUG)
-AC_CXXFLAGS = $(DEBUG)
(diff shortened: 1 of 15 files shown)


wangsheng1001:
The following files were modified based on your local environment:
conf/hostfile
conf/singa.conf
examples/cifar10/job.conf
examples/mnist/job.conf

These should not be committed.

nudles:
@wangsheng1001 @ug93tad I think we may rename this pull request to "Add HDFSStore".
HDFSStore provides functions for reading and writing key-value tuples, which is different to KVFileStore. Hence it does not need to be compatible with KVFileStore (i.e., no need to read the file created from KVFileStore). It it similar for KVFileStore and TextFileStore, which are not compatible.

When users see "Add HDFS support" , they may expect the features as in Hadoop, i.e., uploading their own files and reading it via APIs from HDFS, 
We may support such feature later. For example, we update the KVFile with an option to read data directly from HDFS.  We can also update the TextFileStore to read data directly from HDFS. Then users can upload the files created by KVFileStore/TextFileStore locally to HDFS, and then read these files at runtime via KVFileStore and TextFileStore.

ug93tad:
@wangsheng1001 could you help changing the commit message (at least the 1st line) to reflect the new name?

wangsheng1001:
@ug93tad could you please add a function in create_bin.cc to generate HDFS store data.
Users can create data by sth like: 'make create_hdfs'
These data only need to be created locally, and named: train_data_hdfs,bin, test_data_hdfs.bin.
Let users to put them to their own hdfs is ok.

Please note please do not depend on job.conf when you generate data.

ug93tad:
There'd be more substantial changes to write to get HDFSFile to write to local disk before uploading. I'll get rid of job.conf dependency, but make create_hdfs will upload to HDFS directly.

ug93tad:
@wangsheng1001 please help check the latest commit. You can do "make create_hdfs HDFS_DIR=xxx". The default HDFS_DIR is "hdfs://node0:9000/examples/cifar10".

wangsheng1001:
@ug93tad mnist/create_data.cc does not work, please help check it.

ug93tad:
@wangsheng1001 It should work now (check the commit log). mnist create_data.cc expects different input parameters than cifar10. Use HDFS_MNIST_TRAIN and HDFS_MNIST_TEST to specify the paths (defaults are hdfs://node0:9000/examples/mnist/train_data.bin/test_data.bin)

wangsheng1001:
It seems that create_data.cc still has compile error:
create_data.cc:82:27: error: ‘output_folder’ was not declared in this scope

ug93tad:
My bad. Should compile now.

ug93tad:
wait! there's still a bug with MNIST. let me fix it first.

ug93tad:
Tested with Docker, should run without error now.

PR #98, opened 2015-11-30
SINGA-105 Update GUN make files for compiling cuda related code

xiezl:
To compile with cuda library, try:
./configure --enable-cuda
It searches the system path for cuda header and libs. If cuda is installed in other places, then
./configure --enable-cuda --with-cuda="path to cuda folder"
Do not forget to add those paths to environment variables before run configure script.
If -eanble-cuda or -with-cuda is added, then -DUSE_GPU will be added to the compilation flags.


xiezl:
Merged in #105 .

PR #100, opened 2015-12-04
Feature datavector merge

lzjpaul:
data vector implemented

### DIFF
--- a/include/singa/neuralnet/layer.h
+++ b/include/singa/neuralnet/layer.h
@@ -178,16 +178,38 @@ class Layer {
    * more than one data Blob. In this case, this argument identifies the layer
    * that is requesting the data Blob.
    * @return a const ref for Blob storing feature values of this layer.
+   * @deprecated {This function will be deleted, use
+   * virtual const Blob<float>& data(int k) const instead}.
    */
   virtual const Blob<float>& data(const Layer* from) const {
+    return data_.at(0);
+  }
+  /**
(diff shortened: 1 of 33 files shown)


nudles:
the commit chain is not correct. there should only have commit msgs with title SINGA-110 ...
but your commit log has commits from @wangsheng1001 
pls follow the order exactly,
1. work on your own branch
2. rebase to latest master
3. send push request.

include/singa/neuralnet/layer.h
```diff
@@ -214,8 +236,9 @@ class Layer {
 
  protected:
   LayerProto layer_conf_;
-  Blob<float> data_, grad_;
+  Blob<float> grad_;
```
nudles: type of grad_ is also vector<Blob<float>>.

include/singa/neuralnet/neuron_layer/rbm.h
```diff
@@ -34,11 +34,14 @@ class RBMLayer: virtual public Layer {
  public:
   virtual ~RBMLayer() {}
   void Setup(const LayerProto& proto, const vector<Layer*>& srclayers) override;
-  const Blob<float>& neg_data(const Layer* layer) {
-    return neg_data_;
+  const vector<Blob<float>>& data() const override{
```
nudles: this func should be deleted as it is the same in Layer class.

nudles:
this pull request is replaced with #100, hence it should be closed.

PR #103, opened 2015-12-07
SINGA-110 Add Layer member datavec_ and gradvec_

lzjpaul:
* update Blob<float>& data(const Layer* from) function and Blob<float>* mutable_data(const Layer* from) in layer.h
* update data_, data() function and mutable_data() function in every layer.cc

### DIFF
--- a/include/singa/neuralnet/neuron_layer/rbm.h
+++ b/include/singa/neuralnet/neuron_layer/rbm.h
@@ -34,12 +34,6 @@ class RBMLayer: virtual public Layer {
  public:
   virtual ~RBMLayer() {}
   void Setup(const LayerProto& proto, const vector<Layer*>& srclayers) override;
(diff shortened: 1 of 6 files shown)


include/singa/neuralnet/neuron_layer/rbm.h
```diff
@@ -56,7 +50,7 @@ class RBMLayer: virtual public Layer {
   int batchsize_;
   bool first_gibbs_;
   Param* weight_, *bias_;
-
+  Blob<float> pos_data_;
   Blob<float> neg_data_;
   Blob<float> neg_sample_;
   Blob<float> sample_;
```
nudles: replace pos_sample_ with sample_?

nudles:
may fix the bug from https://issues.apache.org/jira/browse/SINGA-107 in this branch.
just commit in this branch but with the title SINGA-107...

PR #106, opened 2015-12-11
SINGA-81 Add Python Helper, which enables users to construct a model …

chonho:
…(JobProto) and run Singa in Python

- Add Python codes for python users to
    (1) construct a model (JobProto), and
    (2) run singa

- Users are supposed to generate 'usermodel.py'
    * examples are provided, e.g., cifar10_cnn.py, mnist_mlp.py, mnist_rbm.py
    * 'cluster.conf' is required to maintain cluster information

- Users are supposed to run it as follows. e.g.,
    {code}
    cd SINGA_ROOT
    bin/singa-run.sh -conf tool/python/examples/cluster.conf -exe tool/python/examples/mnist_mlp.py
    {code}

- Note: in 'job.proto', required rule of the following fields should be changed to optional
    JobProto: name, neuralnet, train_one_batch, updater, train_steps
    ClusterProto: workspace

- Update README.md
- Add dataset files for the examples


chonho:
still updating README.md

nudles:
@wangsheng1001 

1. we may need to update `singa-run.sh` and zookeeper related code to consider the special case of running on a single node.  For single node training/test, we can directly use the python code to launch program. In other words, there is no need to launch the program via `./bin/singa-run.sh`.

2. for multi-node training, we may need to launch the program as 
 
    ./bin/singa-run/sh -exec user_main.py -conf cluster.conf

  The python code, i.e., user_main.py, would create the JobProto object and pass it to Driver::Train. But ./bin/singa-run.sh needs to get the cluster topology, hence we still need to pass a cluster.conf to it. The cluster.conf has the configuration for a JobProto with all other fields empty except the `cluster` field.

chonho:
For multi-node training, to skip parsing other proto except cluster
we need to change "rule" of some fields in JobProto from "required" to "optional".
e.g., 
JobProto: name, neuralnet, train_one_batch, updater, train_steps
ClusterProto: workspace

tool/python/singa/model.py
```diff
@@ -0,0 +1,362 @@
...
+
+    setval(self.jobconf, train_steps=train_steps)
+    setval(self.jobconf, **fields)
+
+    # loading checkpoint if it is set
+    if data.checkpoint != None:
+      setval(self.jobconf, checkpoint_path=data.checkpoint)
+    
+    # set Train_one_batch component, using backprogapation at default
+    setval(self.jobconf, train_one_batch=Algorithm(type=enumAlgType(alg)).proto)
```
nudles: fit() does not call Driver::Train()?
only evaluate calls Driver::Train().
If user_model.py does not have evaluate(), then the training would not be started.

there are three datasets involved in DL training, training dataset, validation dataset and test dataset.
fit() runs against training (and validation) dataset, whereas evaluate runs against test dataset.
fit() should be independently to evaluate().

Here, we may not be able to follow keras API exactly.

nudles:
I got this error. May be caused by path mismatching?

wangwei@singa1:~/incubator-singa$ ./bin/singa-run.sh -exec tool/python/examples/cifar10_cnn.py -conf cluster.conf 
Unique JOB_ID is 158
Record job information to /tmp/singa-log/job-info/job-158-20151212-143629
Executing : tool/python/examples/cifar10_cnn.py -conf /home/wangwei/incubator-singa/cluster.conf -singa_conf /home/wangwei/incubator-singa/conf/singa.conf -singa_job 158
  File "tool/python/examples/cifar10_cnn.py", line 4, in <module>
    from singa.model import *
  File "tool/python/examples/../singa/model.py", line 3, in <module>
    from layer import *
  File "tool/python/examples/../singa/layer.py", line 2, in <module>
    from parameter import *
  File "tool/python/examples/../singa/parameter.py", line 2, in <module>
    from utils.utility import * 
ImportError: No module named utils.utility
E1212 14:36:31.303450 21582 cluster_rt.cc:419] job 158 not exists

tool/python/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+* Model class has `jobconf` (JobProto) and `layers` (layer list)
+
+Methods in Model class
+
+* add
+	* add Layer into Model
+
+* compile	
+	* set Updater (i.e., optimizer) and Cluster (i.e., topology) components
```
nudles: since cluster topology is defined inside cluster.conf, may comment this out temporarily (I mean comment it in the above line).

tool/python/README.md
```diff
@@ -0,0 +1,305 @@
...
+par = Parameter(init='uniform', range=0.05)
+m.add(Dense(2500, w_param=par, b_param=par, activation='tanh'))
+m.add(Dense(2000, w_param=par, b_param=par, activation='tanh'))
+m.add(Dense(1500, w_param=par, b_param=par, activation='tanh'))
+m.add(Dense(1000, w_param=par, b_param=par, activation='tanh'))
+m.add(Dense(500, w_param=par, b_param=par, activation='tanh'))
+m.add(Dense(10, w_param=par, b_param=par, activation='softmax'))
+
+sgd = SGD(lr=0.001, lr_type='step')
+topo = Cluster(workspace)
```
nudles: comment this line ?

tool/python/examples/mnist_rbm1.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
+import sys, os
+sys.path.append(os.path.join(os.path.dirname(__file__),'..')) 
+from singa.model import *
+from singa.datasets import rbm 
+
+rbmid = 1
+X_train, X_test, workspace = rbm.load_data(rbmid)
+
+m = Sequential('rbm'+str(rbmid), sys.argv)
```
nudles: rbm should not be a sequential model, because the the two rbm layers' connection is undirected.

tool/python/singa/parameter.py
```diff
@@ -0,0 +1,80 @@
...
+        pg = Message('ParamGen', type=enumInitMethod(kwargs['init']), **kwargs)
+        del kwargs['init']
+        setval(self.param, init=pg.proto)
+    else: # default: uniform
+      pg = Message('ParamGen', type=enumInitMethod('uniform'))
+      setval(self.param, init=pg.proto)
+
+  def update(self, **fields):
+    '''
+    update parameter value
```
nudles: for init parameter values?

nudles:
the last comment is on the parameter.py.
i think it is consistent with singa param configuration.
but the python functions are not consistent with keras.
i think it would be better to be consistent with keras *as much as possible*.
we can provide parameter class, but also provide layer configuration following keras.
e.g.,  we assume bias is init with constant value 0, and weight matrix is init with gaussian distribution (0, 0.01).

tool/python/README.md
chonho: currently, we use 'cluster.conf' for cluster topology, but I still expect users to set 'workspace'.
So, I will comment it out now, and modify it. OK?

tool/python/examples/mnist_rbm1.py
chonho: instead of Sequential, Can I call/create "Bidirectional", inherited from Model class?

tool/python/singa/parameter.py
chonho: Ok, I will set parameter initial values to be consistent with Keras
But, after doing that, I think we need to explicitly set the value of most parameters to be consistent with Singa examples.
Anyway, I will update it as soon as possible, and let's discuss this again.

tool/python/examples/mnist_rbm1.py
nudles: Ok. Or call it energy model?

tool/python/singa/parameter.py
nudles: Yes. We try to be consistent with Keira's. The we add arguments for supporting singa features.

tool/python/README.md
```diff
@@ -0,0 +1,330 @@
...
+out_dim = [1000, 500, 250, 30]
+
+for i in range(1, rbmid):
+  m.add(Encoder(i, out_dim[i-1], activation='sigmoid'))
+m.add(Encoder(rbmid, out_dim[-1]))
+
+for i in range(rbmid, 1, -1):
+  m.add(Decoder(i, out_dim[i-2], activation='sigmoid'))
+m.add(Decoder(1, 784, activation='sigmoid'))
+
```
nudles: keras may not provide methods for configuring parameter sharing.
if we want to share parameters between encoders and decoders, we may do it implicitly or explicitly as
1. m.set_param_share(true)
2. m.share(encoder[i], decoder[j])

tool/python/README.md
```diff
@@ -0,0 +1,330 @@
...
+rbmid = 3                                                                                           
+X_train, X_test, workspace = rbm.load_data(rbmid)                                               
+                                                                                              
+m = Energy('rbm'+str(rbmid), sys.argv)
+
+out_dim = [1000, 500, 250]
+m.add(RBM(out_dim, w_std=0.1, b_wd=0)) 
+                                                                                         
+sgd = SGD(lr=0.1, decay=0.0002, momentum=0.8)                                
+topo = Cluster(workspace)
```
nudles: do we need this field `topo`?
if not, we can omit this line.
you can change the `required` to `optional` for some top level configuration fields in src/proto/job.proto. 
then we invoke the program as
`./bin/singa-run./sh -exec tools/python/cifar_cnn.py -conf cluster.conf`
inside cluster.conf, there is only one field `cluster {}`.
in fact, cluster.conf is not necessary for single node training, which would be fixed in a separate ticket later.

tool/python/README.md
```diff
@@ -0,0 +1,330 @@
...
+```
+
+(3) Run singa for test while training 
+```
+m.fit(X_train, nb_epoch=1000, with_test=True)
+result = m.evaluate(X_test, batch_size=100, test_steps=100)
+```
+
+(4) Run singa for test only
+Assume a checkpoint exists after training
```
nudles: there is a function Driver::Test, which reads job conf and does test only.
if we have changed `required` fields to `optional`, we can configure the net structure and test data only to do the test via python scripts, i.e., no need to use `m.fit()`.

tool/python/singa/datasets/ae.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
+from singa.model import *
```
nudles: the auto-encoder example uses the mnist dataset, hence no need to have a separate script for creating dataset?

tool/python/singa/utils/message.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python 
+import sys, os 
+from utility import *
```
nudles: may need some comments for this file at the beginning?

nudles:
1. Pls update the `required` and `optional` fields in job.proto. 
2. rebase -i to clean the commits msg and format the commit title
3. rebase onto the latest master
thx.

nudles:
@chonho could you update this pull request to add support for cudnn layers.
Since keras does not let users to configure the neural net using cudnn layers, we may follow it style:
1. let users configure the gpu device, e.g., model.fit(... device=gpu_id_list)
2. replace the xxxLayer with the corresponding cudnnxxxLayer implicitly.

tool/python/singa/layer.py
```diff
@@ -0,0 +1,431 @@
...
+            setval(self.layer, store_conf=conf.proto)
+
+        self.checkpoint = checkpoint # checkpoint for training data
+
+
+class Convolution2D(Layer):
+
+    def __init__(self, nb_filter=0, kernel=0, stride=1, pad=0,
+                 init=None, w_param=None, b_param=None,
+                 activation=None, **kwargs):
```
nudles: the w_std, and b_lr used in the cifar10 example are not defined here?
users may not know how to use these two arguments.
it would be better to put them in the argument list.

tool/python/singa/layer.py
```diff
@@ -0,0 +1,431 @@
...
+        optional
+          activation = (string)
+          init       = (string)     // 'unirom', 'gaussian', 'constant'
+          w_param    = (Parameter)  // Parameter object for weight
+          b_param    = (Parameter)  // Parameter object for bias
+          **kwargs
+            w_lr = (float) // learning rate for w
+            w_wd = (float) // weight decay for w
+            b_lr = (float) // learning rate for b
+            b_wd = (float) // weight decay for b
```
nudles: xx_lr is the learning rate multiplier for xx, which is used to scale the learning rate when updating parameters.
xx_wd is the weight decay multiplier for xx.

tool/python/singa/layer.py
```diff
@@ -0,0 +1,431 @@
...
+        self.hid_dim = [hid_dim] if type(hid_dim) == int else hid_dim
+
+        self.layer_type = 'AutoEncoder'
+        self.activation = activation
+        self.param_share = param_share
+
+class RBM(Layer):
+
+    def __init__(self, out_dim=None, w_param=None, b_param=None,
+                 sampling=None, **kwargs):
```
nudles: similar as Dense and Convolution2D, the w_std, b_lr .. arguments are not included in the argument list nor explained in the comments.

tool/python/singa/layer.py
chonho: I briefly explain about w_xx and b_xx (e.g w_std, w_lr)in README.
I will explicitly put the message in codes as well

tool/python/singa/layer.py
chonho: I will update it soon

nudles: Thank you!
Just be careful about xx_lr and xx_wd, which are different to lr and wd.
the first set is multiplier, which scales the lr/wd.

chonho:
merged into master

PR #107, opened 2015-12-12
Implementation of GRU layer as well as its test

fanju1984:


include/singa/neuralnet/neuron_layer/gru.h
```diff
@@ -0,0 +1,60 @@
...
+    std::vector<Param*> params{weight_z_hx_, weight_z_hh_, bias_z_,
+                               weight_r_hx_, weight_r_hh_, bias_r_,
+                               weight_c_hx_, weight_c_hh_, bias_c_};
+    return params;
+  }
+
+  int vdim() const { return vdim_;}
+  int hdim() const { return hdim_;}
+  int batchsize() const {return batchsize_;}
+
```
nudles: will the above three functions visible outside of GRU class?
if not, then no need to define them.

src/neuralnet/neuron_layer/gru.cc
```diff
@@ -0,0 +1,265 @@
...
+	CHECK_LE(srclayers.size(), 2);
+
+	// Do transpose
+	Blob<float> *w_z_hx_t = Transpose (weight_z_hx_->data());
+	Blob<float> *w_z_hh_t = Transpose (weight_z_hh_->data());
+	Blob<float> *w_r_hx_t = Transpose (weight_r_hx_->data());
+	Blob<float> *w_r_hh_t = Transpose (weight_r_hh_->data());
+	Blob<float> *w_c_hx_t = Transpose (weight_c_hx_->data());
+	Blob<float> *w_c_hh_t = Transpose (weight_c_hh_->data());
+
```
nudles: there will be a function in Blob::T() which generates the transpose Blob in the master.
then, no need to explicitly define these pointers. some pointers are used only once. but we have to delete them...

src/neuralnet/neuron_layer/gru.cc
```diff
@@ -0,0 +1,265 @@
...
+  hdim_ = layer_conf_.gru_conf().dim_hidden(); // dimension of hidden state
+
+  data_.Reshape(vector<int>{batchsize_, hdim_});
+  grad_.ReshapeLike(data_);
+
+  // Initialize the parameters
+  weight_z_hx_ = Param::Create(conf.param(0));
+  weight_z_hh_ = Param::Create(conf.param(1));
+  bias_z_ = Param::Create(conf.param(2));
```
nudles: some gru do not have bias? 
hence may need to add a configuration field, e.g., has_bias.
if it is true, then new the bias param.

src/neuralnet/neuron_layer/gru.cc
```diff
@@ -0,0 +1,265 @@
...
+	GEMM(cpu, 1.0f, 0.0f, src,*w_z_hx_t,update_gate);
+	MVAddRow(cpu,1.0f,1.0f,bias_z_->data(),update_gate);
+	Blob<float> zprev (batchsize_,hdim_);
+	GEMM(cpu, 1.0f, 0.0f, *context,*w_z_hh_t, &zprev);
+	Add<float>(cpu, *update_gate, zprev, update_gate);
+	Map<op::Sigmoid<float>,float>(cpu, *update_gate, update_gate);
+
+	// Compute the reset gate
+	GEMM(cpu, 1.0f, 0.0f, src,*w_r_hx_t,reset_gate);
+	MVAddRow(cpu,1.0f,1.0f,bias_r_->data(),reset_gate);
```
nudles: check bias==nullptr before MVAddRow.

src/neuralnet/neuron_layer/gru.cc
```diff
@@ -0,0 +1,265 @@
...
+	Blob<float> rprev (batchsize_, hdim_);
+	GEMM(cpu, 1.0f, 0.0f, *context, *w_r_hh_t, &rprev);
+	Add<float>(cpu, *reset_gate, rprev, reset_gate);
```
nudles: line 133 and 134 can be merged into one line
GEMM(cpu, 1.0f, 1.0f, *context,*w_z_hh_t, update_gate) ?

src/test/gru-in-1.csv
```diff
@@ -0,0 +1,2 @@
+0,0,0,1
```
nudles: only .cc and .h files should be tracked by git.

nudles:
I just went through the ComputeFeature function. I think there should be no bugs. Minor changes are suggested.
Do you have links to webpages that shows the compute gradient equations?
Thanks.

fanju1984:
Regarding your question, "Do you have links to webpages that show the equations for gradient computation?", I did not find web pages for gradient computation... 

I computed the gradients by myself. Do you know if there is any relevant website? then I can check.

include/singa/neuralnet/neuron_layer/gru.h
fanju1984: I think no.

src/neuralnet/neuron_layer/gru.cc
fanju1984: Can you provide more explanation of Blob::T()? Does it create a new blob?

src/neuralnet/neuron_layer/gru.cc
fanju1984: OK. I will refine it.

src/neuralnet/neuron_layer/gru.cc
fanju1984: Yes, you are right.

src/test/gru-in-1.csv
fanju1984: If I want to exclude the csv files, should I add *.csv into git-ignore?

src/neuralnet/neuron_layer/gru.cc
fanju1984: By the way, you mentioned "there will be". When this new function will be added?

src/neuralnet/neuron_layer/gru.cc
fanju1984: I checked this again and find it may not be merged, as the elements in "update_gate" should be initially set to 0 - that is why I used GEMM(cpu,1.0f,0.0f,....)...

nudles: I see.

src/neuralnet/neuron_layer/gru.cc
nudles: it will be available after SINGA-100 is merged.

nudles:
I don't have...

src/neuralnet/neuralnet.cc
```diff
@@ -99,11 +107,89 @@ NeuralNet* NeuralNet::Create(const NetProto& net_conf, Phase phase,
...
   LOG(INFO) << "NeuralNet config is\n" << conf.DebugString();
   // TODO(wangwei) create net based on net type, e.g., directed, undirected, etc
   return new NeuralNet(conf, npartitions);
 }
 
+NetProto NeuralNet::Unrolling(const NetProto& net_conf) {
+	// Step 1: Unroll each layer & set parameter sharing
+	NetProto conf;
+	conf.CopyFrom(net_conf);
+	conf.Clear();
```
nudles: CopyFrom and then Clear？
can we remove 119 and 120?

src/neuralnet/neuralnet.cc
```diff
@@ -99,11 +107,89 @@ NeuralNet* NeuralNet::Create(const NetProto& net_conf, Phase phase,
...
+			layer_conf->clear_unroll_len();
+			layer_conf->clear_unroll_conn_type();
+			layer_conf->clear_shift();
+			layer_conf->clear_srclayers();
+
+			layer_group.push_back(conf.layer_size() - 1);
+		}
+		layer_groups.push_back(layer_group);
+	}
+	// Step 2: Connect unrolled layers by setting `srclayers`
```
nudles: may better differentiate the original layer (and its srclayers) with unrolled layer (and srclayers) using different variable names, just for easy to read:)

fanju1984: Hi, I have updated the variable names. Are they better now?

nudles: thanks. I can help update the code if you add me as your collaborator. but my network is not stable these two days..

src/test/test_unrolling.cc
```diff
@@ -0,0 +1,107 @@
...
+	driver.RegisterLayer<SoftmaxLossLayer, int>(kSoftmaxLoss);
+	driver.RegisterLayer<EuclideanLossLayer,int>(kEuclideanLoss);
+
+
+
+	JobProto job;
+	job.CopyFrom(job_conf0);
+	cout << "Create Train Net" << endl;
+	NeuralNet* train_net = NeuralNet::Create(job.neuralnet(), kTrain, 1);
+	cout << "# of layers in Train Net: " << train_net->layers().size();
```
nudles: can replace cout with assert_eq or expect_eq.
gtest uses such macros for testing.

typically gtest does not depend on static files, as gru-unroll-1.conf, because gtest is unit test, which should only depends on the functions being tested.
like test_gru.cc, we can generate the JobProto and then use as input for testing functions/classes, and the use macros to test the correctness using outputs from the functions/classes.
this part can be updated later:)
thanks.

fanju1984: Hi Wei, Thanks for the useful information. I will update the tests in these days.

src/neuralnet/neuralnet.cc
fanju1984: Added. Thanks.

Makefile.am
```diff
@@ -272,8 +274,22 @@ singatest_LDFLAGS = -lsinga \
...
+endif
+
+_driver_la_SOURCES = $(PY_SRCS)
+_driver_la_CXXFLAGS = $(DEFAULT_FLAGS) $(MSHADOW_FLAGS) -I$(top_srcdir)/include $(PYFLAGS)
+_driver_la_LDFLAGS = -lsinga -module -shared $(PYLIBS) -avoid-version -rpath $(pydir)
+
+if DCUDA
+_driver_la_CXXFLAGS += $(CUDA_CFLAGS)
+_driver_la_LDFLAGS += $(CUDA_LDFLAGS) $(CUDA_LIBS)
+endif
```
nudles: should delete 277-290

src/neuralnet/neuralnet.cc
```diff
@@ -78,16 +77,25 @@ NeuralNet* NeuralNet::Create(const NetProto& net_conf, Phase phase,
...
-        shares.push_back(param);
-    }
+  }
+  //LOG(INFO) << "Before unrolling: \n" << conf.DebugString();
+  conf = Unrolling (conf);
+
+  // Copy shared parameters for sharing param conf
+  std::unordered_map<string, ParamProto*> name2param;
+  std::vector<ParamProto*> shares;
+  for (int index = 0; index < conf.layer_size();index ++) {
```
nudles: need to format this piece of code.

src/test/store.txt
```diff
@@ -0,0 +1,3 @@
+first tuple
+second tuple
```
nudles: delete this file

src/test/test.csv
```diff
@@ -0,0 +1,3 @@
+12,3.2,1,14.1
+2,0.2,0,1.1
```
nudles: delete this file

PR #109, opened 2015-12-22
Logging

cac2003:
add switch for enabling/disabling INFO-level logging


include/singa/log.h
```diff
@@ -0,0 +1,29 @@
...
+* specific language governing permissions and limitations
+* under the License.
+*
+*************************************************************/
+
+#ifndef SINGA_LOG_H_
+#define SINGA_LOG_H_
+
+namespace singa {
+    extern bool enable_info_log;
```
nudles: maybe a shorter name would be better as it is used frequently (easier for typing), e.g., DEBUG, kDebug, verbose_singa.. verbose is typically used as foo(arg1, arg2, verbose).

may put the declaration in singa.h, default value is false. the main function can parse a command line argument for this variable, e.g., -verbose true/false.

src/comm/socket.cc
```diff
@@ -120,7 +121,7 @@ int Router::Bind(const std::string& endpoint) {
     port = zsock_bind(router_, "%s", endpoint.c_str());
   }
   CHECK_NE(port, -1) << endpoint;
-  LOG(INFO) << "bind successfully to " << zsock_endpoint(router_);
+  LOG_IF(INFO, enable_info_log) << "bind successfully to " << zsock_endpoint(router_);
```
nudles: We were not consistent with the Log msg format.
Maybe uniform them to start with a capital word, e.g., "Bind..."

include/singa/log.h
cac2003: Will replace it with verbose_singa.

If its declaration is put in singa.h, then every src file where this var appears should include singa.h as well as the files included in singa.h. Is it a bit redundant? 

Since this is a global variable and defined as external, its real declaration and initialization should be put somewhere else.  Which file do you think is appropriate for it?

btw, where is the main entrance(function) for singa located, singa/main.cc or utils/tool.cc?

nudles: singa_verbose, Verbose, verbose_singa, which one is better?

singa.h is not included/used internally. it is to be included by user code.
hence other files inside src/ do not need to include singa.h to get this
external variable, right?

the main entrance is src/main.cc

cac2003: 1. singa_verbose?

2. I am afraid not. If this external var is defined in singa.h, then other files where this var appears needs either to include singa.h, or to have this external variable defined in its header. (The latter seems more feasible)

3. So we can place the real definition and initialization in src/main.cc, as long as it is compiled into singa executable.

nudles:
can we do it in this way?

in singa.h

    namespace singa{
      external bool singa_verbose;
    }

in main.cc

    #include "singa.h"
    ...
    if -verbose in argv
       singa_verbose = true 
    else 
       singa_verbose = false

nudles:
@cac2003 I am going to merge this pull request. before that, pls clean the commit msgs following the format http://singa.apache.org/develop/contribute-code.html.
1. the commit msg title should be SINGA-xxx aaaaaaaaaa
2. squash small commits together using `git rebase -i commit_id`
3. rebase to the latest master

cac2003:
how to determine the xxx part of commit msg title?

nudles:
Once you create the jira ticket you will see a ticket No., which is the xxx

nudles:
Could you pls uniform the format of all key-value tuples in the log?
E.g., key = value[, key = value]  (or other format that is easy to read).
It would be useful (easy) to parsing the log file to get all key-value tuples.
Thanks.

PR #110, opened 2015-12-24
SINGA-115 - Print layer debug information in the neural net graph file

nudles:
* Updated the Graph class to format the string representation in ToJson function.
* The Layer::ToString function is updated to generate string for debug info, e.g., layer data and gradient norm.
* Insert code to collect debug info from layers forward and backward function in BPWorker class.
* Add NeuralNet::ToGraph function to convert a neualnet into a Graph which can converted into a string using ToJson function.

To test this pull request,

1. run the cifar10 example, get the `train_net.json` file from `example/cifar10/visualization/`, visualize the neural net by 

        python tool/graph.py examples/cifar10/visualization/train_net.json train_net.jpg

2. run the cifar10 example for a few steps without test and with debug enabled. 
      
        // job.conf 
        train_steps: 5
        test_steps: 0
        debug: true

  Visualize the neural net with debug info recorded in `examples/cifar10/visualization/*.json` in the same way as 1.

### DIFF
--- a/include/singa/neuralnet/connection_layer.h
+++ b/include/singa/neuralnet/connection_layer.h
@@ -97,6 +97,7 @@ class SliceLayer : public ConnectionLayer {
   void Setup(const LayerProto& proto, const vector<Layer*>& srclayers) override;
   void ComputeFeature(int flag, const vector<Layer*>& srclayers) override;
   void ComputeGradient(int flag, const vector<Layer*>& srclayers) override;
+  const std::string ToString(bool debug, int flag) override;
   const Blob<float>& data(const Layer* from) const override;
   const Blob<float>& grad(const Layer* from) const override;
   Blob<float>* mutable_data(const Layer* from) override;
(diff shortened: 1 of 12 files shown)


lijiansong:
Why this tool is removed in the later version of Singa? I think this tool might be very useful.

XJDKC:
> Why this tool is removed in the later version of Singa? I think this tool might be very useful.
> 
> Is there any way to visualize the computation graph in the latest version of Singa?

Hi, I'm sorry I don't know why we remove this tool but I know how to visualize the graph. There is a function in [scheduler.cc](https://github.com/apache/singa/blob/dev/src/core/scheduler/scheduler.cc#L116) which can be used to print the graph info. If we want to visualize the graph, we can use this function to obtain the structure info of the neural network and then use some python library(e.g. matplotlib and networkx) to achieve that.

lijiansong:
@XJDKC  Okay, thanks for your reply.

PR #112, opened 2015-12-29
SINGA-114, 118 improvement on log dir and layer type id

wangsheng1001:
We init log dir before calling cluster setup, so that it will not generate a short log in tmp dir.

clean ids in job.conf:
     - Input Layer:      100 - 199
     - Neuron Layer:     200 - 299
     - Loss Layer:       300 - 399
     - Output Layer:     400 - 499
     - Connection Layer: 500 - 599

### DIFF
--- a/src/proto/job.proto
+++ b/src/proto/job.proto
@@ -21,22 +21,24 @@
 
 package singa;
 
-// To start a training job, all we need is a JobProto object.
-// It should contain following fields
-//  - Job Name (name)
-//      the name to identify the job
-//  - NeuralNet (neuralnet)
-//      the neural network structure contains a set of layers
-//  - Train One Batch (alg)
-//      the training algorithm
-//  - Updater (updater)
-//      the protocol for updating parameters at server side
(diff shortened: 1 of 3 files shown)


src/proto/job.proto
```diff
@@ -21,22 +21,24 @@
...
+ *  - Cluster Topology (cluster)
+ *      the distributed topology of workers/servers
+ *  - Training Steps (train_steps)
+ *      the number of training iteration
+ *  All other fields/functions are optional, e.g., test, checkpoint
+ */
+
 message JobProto {
   // job name, e.g., "cifar10-dcnn", "mnist-mlp"
   required string name = 1;
```
nudles: I am going to replace the `required` fields with `optional`.
The python binding will configure most fields using a python script except the cluster topology.
Hence we invoke the program like `./bin/singa-run.sh -exec myscript.py -conf cluster.conf`
the cluster.conf contains only cluster topology fields, hence need to make other fields optional.

PR #119, opened 2016-01-07
SINGA-125 Improve Python Helper

chonho:
- Update README.md

- Update layer.py and model.py
  . deal with non-square values for kernel, stride, pad
  . users can specify Accuracy layer by 'show_acc=True'

- Update cifar10 examples
  . set momentum to 0.9 to speed up convergence

### DIFF
--- a/tool/python/README.md
+++ b/tool/python/README.md
@@ -1,163 +1,77 @@
-# Python Helper
+# Python Binding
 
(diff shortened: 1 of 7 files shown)


nudles:
chonho, could you please update the license header in __init__.py files. I added the licenses with the c style comments...
thanks.

PR #125, opened 2016-01-18
SINGA-125 Improve Python Helper

chonho:
- Update model.py to get TANH for gpu when activation='tanh'. (It was STANH)


tool/python/singa/model.py
```diff
@@ -325,7 +325,7 @@ def set_cudnn_layer_type(self, net):
             elif ly_type == kSoftmaxLoss: cudnn_ly_type = kCudnnSoftmaxLoss
             elif ly_type == kSTanh:
                 cudnn_ly_type = kCudnnActivation
-                net.layer[i].activation_conf.type = STANH
+                net.layer[i].activation_conf.type = TANH
```
nudles: STANH is not available in cudNN.
Hence if users configure stanh, then just report error?

chonho: Agree,
for GPU
I keep activation="tanh" to TANH, and I will report error for activation="stahn"

By the way, for CPU
I changed  activation="tanh" to STANH, and I did NOT consider TANH.
Should I change to action="tanh" to TANH  and  activation="stanh" to STANH ?

nudles: Currently, STANH is not enabled for CPU, hence pls just report an error when users configure STANH for the type of activation layer.

tool/python/singa/layer.py
```diff
@@ -231,19 +231,23 @@ class Activation(Layer):
...
+          print 'Warning: Tanh layer is not supported for CPU'
+
         self.name = activation
-        if activation == 'tanh': activation = 'stanh' # <-- better way to set?
         self.layer_type = enumLayerType(activation)
         super(Activation, self).__init__(name=generate_name(self.name),
                                          type=self.layer_type)
+
         if activation == 'softmaxloss':
             self.layer.softmaxloss_conf.topk = topk
```
nudles: Is softmaxloss an activation layer in Keras?

chonho: Keras set softmaxloss layer by
model.compile(loss='categorical_crossentropy')

nudles: then activation could never be 'softmaxloss', right?

chonho: no in keras
I just keep it in singa for users setting either way.

nudles: Activation function could be relu, sigmoid, tanh, stanh and softmax. But softmaxloss is a loss function. Hence it may confuse people if we use it as an activation function.

chonho: ok. I will take care of it and make it clear to users.

tool/python/singa/layer.py
```diff
@@ -228,21 +228,20 @@ def __init__(self, size=0, **kwargs):
...
+    def __init__(self, activation='stanh'):
         '''
         required
-          activation = (string)
-        optional
-          topk       = (int)  // the number of results
+          activation = (string) // relu, sigmoid, tanh, stanh, softmax.
         '''
 
+        if activation != 'tanh':
```
nudles: Is it activation == 'tanh' or activation != 'stanh'?

chonho: Sorry forgot to change. Fixed.

tool/python/singa/layer.py
```diff
@@ -228,21 +228,20 @@ def __init__(self, size=0, **kwargs):
...
+    def __init__(self, activation='stanh'):
         '''
         required
-          activation = (string)
-        optional
-          topk       = (int)  // the number of results
+          activation = (string) // relu, sigmoid, tanh, stanh, softmax.
         '''
 
+        if activation == 'tanh':
```
nudles: sorry. I think here is still some problem.
currently, ActivationLayer and cudnnActivationLayer support sigmoid, relu and tanh. 
Besides, there is a STanhLayer which supports stanh.
Hence, when users configure the activation to be tanh for cpu or gpu, we can support it.
When users configure the activation to be stanh, we can create the STanhLayer to support it.
Hence, we may not need to warn the users on tanh for cpu. Am I right?

chonho: In short,
- singa supports sigmoid, relu, tanh for both cpu and gpu by ActivationLayer and cudnnActivationLayer, respectively.
- It supports stanh by STanhLayer for cpu.
- We will update it for gpu soon.

Thus, simply we remove lines 237 and 238, right?

tool/python/singa/layer.py
```diff
@@ -228,21 +228,20 @@ def __init__(self, size=0, **kwargs):
...
+          print 'Warning: Tanh layer is not supported for CPU'
+
         self.name = activation
-        if activation == 'tanh': activation = 'stanh' # <-- better way to set?
         self.layer_type = enumLayerType(activation)
```
nudles: Do you process cpu tanh and cpu stanh in the same way?
Currently, tanh can be implemented by creating an ActivationLayer and setting its type to be TANH. STANH is implemented by creating a STanHLayer.
I think line 241 only gets the layer type. line 242 passes it to create the Layer object. 
Where do you set the activation type for tanh?

chonho: For CPU, I thought user need set only layer type.
If it requires to set activation_conf as well, then I will fix it

nudles: Currently, there is a SigmoidLayer, STanhLayer and ReLULayer, hence you can use them if the activation is sigmoid/stanh/relu.
But for tanh, there is no TanhLayer, hence you need to create an ActivationLayer and set its type to TANH. 

A better approach is to uniform all these activation types. I think we can deprecate ReLULayer/SIGMOIDLayer/STanhLayer. But we need to update the ActivationLayer to enable STANH.

chonho: Can I add "kActivation" enum in LayerType for tanh/relu/sigmoid ?

Then, I will set the ActivationLayer type to TANH/ReLU/SIGMOID
In this way, it will be clearer to distinguish cpu and gpu

Later, stanh will follow the same manner after updating ActivationLayer to enable STANH

chonho: Can I add "kActivation" enum in LayerType for tanh/relu/sigmoid ?
Then, I will set the ActivationLayer type to TANH/ReLU/SIGMOID
In this way, it will be clearer to distinguish cpu and gpu

Later, stanh will follow the same manner after updating ActivationLayer to enable STANH
(now I keep kSTanh)

If ok, then we need to register "kActivation" for ActivationLayer in driver.cc , right?

nudles: If we use kActivation, then it would be different to keras for the activation layers..
With kActivation, how do you distinguish cpu and gpu?

chonho:
This was solved in SINGA-126.

PR #127, opened 2016-01-27
add clocks for forward and backward functions.

daijian:
add clocks for forward and backward functions.

### DIFF
--- a/src/utils/clock.cc
+++ b/src/utils/clock.cc
@@ -1,0 +1,61 @@
+/************************************************************
+*
+* Licensed to the Apache Software Foundation (ASF) under one
+* or more contributor license agreements.  See the NOTICE file
+* distributed with this work for additional information
+* regarding copyright ownership.  The ASF licenses this file
+* to you under the Apache License, Version 2.0 (the
+* "License"); you may not use this file except in compliance
+* with the License.  You may obtain a copy of the License at
+*
+*   http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*
+*************************************************************/
+
+#include "singa/utils/clock.h"
+
+#include <glog/logging.h>
+#include <cmath>
+#include <ctime>
+#include "singa/utils/cluster.h"
+#include "singa/utils/factory.h"
+#include "singa/utils/singleton.h"
+#include "singa/utils/context.h"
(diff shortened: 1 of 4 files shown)


wangsheng1001:
Hi daijian,

Please kindly follow the guideline below to format your pull request before we can merge it:
http://singa.incubator.apache.org/develop/contribute-code.html

Please include JIRA ticket in your commits and check coding style using cpplint.

src/utils/clock.cc
```diff
@@ -0,0 +1,61 @@
...
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+*
+*************************************************************/
+
+#include "singa/utils/clock.h"
```
nudles: Considering this class is simple, maybe we can move the implementation into the .h file..
Timer vs Clock, which one is better?
c++11 has a set of functions related to time in<chrono>. There could be some functions that are better (more accurate) than **clock**.

src/utils/clock.cc
```diff
@@ -0,0 +1,61 @@
...
+#include <ctime>
+#include "singa/utils/cluster.h"
+#include "singa/utils/factory.h"
+#include "singa/utils/singleton.h"
+#include "singa/utils/context.h"
+
+namespace singa {
+
+void Clock::start() {
+	start_t = clock();
```
nudles: if we create an initiator to start the timer/clock, then we can save one line when using it, e.g.,

    Clock c;
      // computation
    c.elapse();

we still keep the start and end functions. the elapse() function returns the time interval.

src/worker.cc
```diff
@@ -335,6 +336,8 @@ void BPWorker::TestOneBatch(int step, Phase phase, NeuralNet* net) {
 }
 
 void BPWorker::Forward(int step, Phase phase, NeuralNet* net) {
+  Clock c;
+ c.start();
```
nudles: it seems the indent is inconsistent.
pls use cpplint.py to check the format.

src/utils/clock.cc
```diff
@@ -0,0 +1,61 @@
...
+	return runtime;
+}
+
+double Clock::end(int iteration_num, std::string content) {
+ end_t=clock();
+ double runtime=(end_t-start_t)/(double(CLOCKS_PER_SEC)*1000*iteration_num);
+ return runtime;
+}
+
+void Clock::endWithLog( std::string content) {
```
nudles: The function names should start with upper case letters except set/get functions.
Would it be better to use EndWithDLog as you used DLog for logging?

daijian:
SINGA-138 add clocks for forward and backward functions.

PR #128, opened 2016-02-01
Singa-143 The compilation cannot detect libsingagpu.so file

xiezl:
Add the singa folder into ld parameters when compiling.
Also fix SINGA-112


README.md
```diff
@@ -43,6 +43,7 @@ For advanced features, the following libraries are needed:
   * `Apache Mesos` (Apache 2.0)
   * `Apache Hadoop` (Apache 2.0)
   * `libhdfs3` (Apache 2.0)
+  * `swig` (GPL)
```
nudles: It is GPL v3 which is compatible with Apache V2.

PR #130, opened 2016-02-23
SIGNA-141 Undesired Hash collision when locating process id to worker…

ijingo:
…s and servers

Circumvent undesired collision in Hash function by adding maximum range from calculating for workers
as the offset when calculating for servers.


nudles:
Could you pls give one example illustrating the hash collision case?
Thanks.

ijingo:
Let us assume that there are 2 worker groups and 2 workers per group and 1 server in 1 server group, and we separate server and worker in different processes. The original Hash(1,0,0) and Hash(0,0,2) will get the same value 2.

PR #132, opened 2016-03-13
SINGA-148 Race condition between Worker threads and Driver

undisputed-seraphim:
In this PR I aim to fix the race condition between Driver and Worker threads.

Included are the additional *.conf files I created to perform testing under various hardware conditions where other variables remain equal. You can ignore them.

The main changes made are in Driver::Train(), Driver::CreateWorkers(), Worker::Run(), and job.proto.
I have added an additional variable to be passed in during the creation of Worker objects so the Worker knows its identity as a CPU or GPU  execution unit.
Worker objects also perform context->SetupDevice() now instead of deferring it to Driver::Train().
The execution flow in Driver::CreateWorkers() has also been straightened out a bit.

### DIFF
--- a/include/singa/utils/cluster.h
+++ b/include/singa/utils/cluster.h
@@ -53,6 +53,7 @@ class Cluster {
   inline int nservers_per_group() const { return cluster_.nservers_per_group();}
   inline int nworkers_per_procs() const { return cluster_.nworkers_per_procs();}
   inline int nservers_per_procs() const { return cluster_.nservers_per_procs();}
+  inline int ngpu_per_group() const { return cluster_.ngpu_per_group();}
   inline int nworker_groups_per_server_group() const {
     if (nserver_groups() == 0 || nservers_per_group() == 0)
       return 1;
--- a/include/singa/worker.h
+++ b/include/singa/worker.h
@@ -62,6 +62,12 @@ class Worker {
    * @return a pointer to the instance of the Worker subclass.
    */
   static Worker* Create(const AlgProto& conf);
+  /**
+   * Create an instance of the subclass of Worker
+   * with the specified device type.
+   */
+  static Worker* Create(const AlgProto& conf, DeviceType devtype);
+
   virtual ~Worker();
   /**
    * @param[in] grp_id global worker group ID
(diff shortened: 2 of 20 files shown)


include/singa/utils/cluster.h
```diff
@@ -53,6 +53,7 @@ class Cluster {
   inline int nservers_per_group() const { return cluster_.nservers_per_group();}
   inline int nworkers_per_procs() const { return cluster_.nworkers_per_procs();}
   inline int nservers_per_procs() const { return cluster_.nservers_per_procs();}
+  inline int ngpu_per_group() const { return cluster_.ngpu_per_group();}
```
nudles: ->ngpus_per_group (to be consistent with the naming convention).

include/singa/worker.h
```diff
@@ -62,6 +62,12 @@ class Worker {
    * @return a pointer to the instance of the Worker subclass.
    */
   static Worker* Create(const AlgProto& conf);
+  /**
+   * Create an instance of the subclass of Worker
+   * with the specified device type.
+   */
+  static Worker* Create(const AlgProto& conf, int devid);
```
nudles: can we bind the work to the device after creating the threads?
if so, then we do not need this API. 
instead, we can assign worker thread a device ID (gpu or cpu) through Context.
workers can then query Context to know its device IDs.

include/singa/worker.h
```diff
@@ -263,14 +269,26 @@ class Worker {
...
   /**
-   * @reutrn worker ID within the worker group.
+   * @return worker ID within the worker group.
    */
   inline int id() const { return id_; }
 
+  /**
+   * @return The device ID of this worker. It is -1 for CPU and >-1 for GPU.
+   */
+  inline int device_id() { return device_id_; }
```
nudles: device_id can be accessed via Context. 
it would be better to avoid storing the same variable in two places, which would lead to inconsistent if only one place is updated.

src/worker.cc
```diff
@@ -62,12 +65,13 @@ Worker::~Worker() {
 }
 
 void Worker::Run() {
-  // setup gpu device
   auto context = Singleton<Context>::Instance();
+  context->SetupDevice(std::this_thread::get_id(), device_id_);
```
nudles: It is preferred to use thread pool to setup Context (thread_id, device_id) than passing device_id as an argument of creating Worker instance.

nudles:
thanks for sending this pull request. here are some suggestions,
1. pls only commit the changes that are necessary for your features, e.g., many conf files should not be changed (cifar10/job.conf, cifar10/gpu.conf,...)
2. see my in-line comments.
3. after addressing the above comments, pls re-format the commit messages (using rebase to merge small commits, and add the JIRA ticket number at the beginning of the title of each commit message).

nudles:
Pls try #139 to see if it can fix this problem.

PR #133, opened 2016-03-14
 	SINGA-80 New Blob Level and Address Level Math Operation Interface

jinyangturbo:
add gtest for blob level operations
and bug fix in math_blob.h


nudles:
It seems that you worked on an old branch?
You can create a new ticket and add the updates against the latest code.

PR #136, opened 2016-03-18
SINGA-126 Python Binding for Interactive Training

chonho:
- Update tool/python/singa/driver.i for wrapper

- Add DummyInputLayer
  . kDummyInput in job.proto
  . register this layer in driver.cc

- Revise python codes
  . layer.py, model.py, utility.py
  . In layer.py
    . add methods: ComputeFeature, ComputeGradient, GetParams, GetData

- Add example
  . train_mnist.py
  . train_cifar10.py


src/neuralnet/neuralnet.cc
```diff
@@ -58,6 +58,12 @@ const NetProto NetConfPreprocess(const NetProto& conf) {
   return proto;
 }
 
+NeuralNet* NeuralNet::CreateForTest(const string str) {
```
aaronwwf: "CreateFromStr" is a better name.

tool/python/examples/train_cifar10.py
```diff
@@ -0,0 +1,133 @@
...
+def get_labellist():
+    dataset_dir_ = singa_root_ + "/examples/cifar10/cifar-10-batches-bin"
+    fname_label_list = dataset_dir_ + "/batches.meta.txt"
+    label_list_ = np.genfromtxt(fname_label_list, dtype=str)
+    return label_list_
+
+#-------------------------------------------------------------------
+print '[Layer registration/declaration]'
+d = Driver()
+d.Init(sys.argv)
```
aaronwwf: Need to explain the system arguments

tool/python/examples/train_cifar10.py
```diff
@@ -0,0 +1,133 @@
...
+
+#-------------------------------------------------------------------
+print '[Start training]'
+batchsize = 100 
+disp_freq = 50
+train_step = 1000
+
+for dataset_id in range(train_step / batchsize):
+
+    x, y = load_dataset(dataset_id%5+1)
```
aaronwwf: Need to preprocess datasets, now it calculates mean value at every time.

tool/python/singa/layer.py
```diff
@@ -41,13 +45,110 @@ def __init__(self, **kwargs):
...
+        assert upd != None, 'required Updater (see model.py)' 
+        if Layer.singaupdater == None:
+            Layer.singaupdater = SingaUpdater.CreateUpdater(upd.proto.SerializeToString()) 
+
+        # call ComputeGradient of Singa
+        self.singalayer.ComputeGradient(1, self.singaSrclayerVector)
+        SingaUpdater.UpdateParams(Layer.singaupdater, self.singalayer, step)
+
+        # recursively call ComputeGradient of srclayers
+        for sly in self.srclayers:
```
aaronwwf: If a layer has multiple children, the layer's computeGradient method will be called multiple times.

tool/python/examples/train_cifar10.py
```diff
@@ -0,0 +1,136 @@
...
+
+    for i in range(x.shape[0] / batchsize):
+
+        xb, yb = x[i*batchsize:(i+1)*batchsize,:], y[i*batchsize:(i+1)*batchsize,:]
+        input.SetData(xb, 3, 0)
+        label.SetData(yb, 1, 1)
+        
+        conv1.ComputeFeature(input)
```
nudles: can we use a loop to go through all layers?
or add a comment for it.

Makefile.am
```diff
@@ -64,6 +64,7 @@ SINGA_SRCS := src/driver.cc \
               src/neuralnet/input_layer/record.cc \
               src/neuralnet/input_layer/deprecated.cc \
               src/neuralnet/input_layer/store.cc \
+              src/neuralnet/input_layer/dummy.cc \
```
nudles: there is a DummyLayer in src/neuralnet/neuron_layer/dummy.cc.
can we merge these two layers @wangsheng1001 ?
I think their functionality should be similar.

include/singa/neuralnet/layer.h
```diff
@@ -69,6 +70,15 @@ class Layer {
...
   Layer() {}
   virtual ~Layer() {}
+
+  /**
+   * Create for python binding, production test mode
+   *
+   */
+  static Layer* CreateLayer(const string str);
+  static void SetupLayer(Layer* layer, const string str, const vector<Layer*>& srclayers);
+  virtual void Feed(int batchsize, vector<int> shape, vector<float>* data, int op) {}
```
nudles: Do we need a Feed function for every layer? or just the DummyLayer?

tool/python/examples/train_cifar10.py
chonho: possible to do
(i) if user create a model m (like what we do in singa-125), then we can traverse it 
or
(ii) create a list or dict for maintaining a network in [conv1, pool1, act1, ... ] and go through it.

I will work on it tomorrow. pls let me know if you have better suggestion.

Thanks.

include/singa/utils/updater.h
```diff
@@ -87,6 +90,12 @@ class InvTLRGen : public LRGenerator {
  */
 class Updater {
  public:
+
+  /* added for python binding */
+  static Updater* CreateUpdater(const string str);
+  static void UpdateParams(Updater* updater, Layer* layer, int step);
```
nudles: Do we have a python class for Param?
If yes, then we do not need the UpdateParams function..

include/singa/neuralnet/layer.h
chonho: oh, good point. I will check and fix it.

include/singa/utils/updater.h
```diff
@@ -93,7 +93,6 @@ class Updater {
 
   /* added for python binding */
   static Updater* CreateUpdater(const string str);
-  static void UpdateParams(Updater* updater, Layer* layer, int step);
   /* ------------------------ */
 
   static Updater* Create(const UpdaterProto& proto);
```
nudles: CreateUpdaterFromStr (other functions are named as xxxFromStr?)

tool/python/examples/train_cifar10_2.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python
```
nudles: can we put the loop code as comments in train_cifar10.py (most of the code is the same as that)?

tool/python/examples/train_cifar10_2.py
```diff
@@ -0,0 +1,123 @@
...
+
+    x, y = load_dataset(dataset_id%5+1)
+
+    for i in range(x.shape[0] / batchsize):
+
+        xb, yb = x[i*batchsize:(i+1)*batchsize,:], y[i*batchsize:(i+1)*batchsize,:]
+        input.SetData(xb, 3, 0)
+        label.SetData(yb, 1, 1)
+        
+        for h in range(len(nn)):
```
nudles: how about
inpu.feed()
for i in range(1, len(nn)):
  nn[i].computefeature()

tool/python/examples/train_mnist_2.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python
```
nudles: similarly, we can replace the code 105-115 in train_minst.py with the following loop or add the following code as comments?

src/neuralnet/neuron_layer/dummy.cc
```diff
@@ -71,4 +78,49 @@ void DummyLayer::ComputeGradient(int flag, const vector<Layer*>& srclayers) {
     Copy(grad_, srclayers[0]->mutable_grad(this));
 }
 
+void DummyLayer::Feed(vector<int> shape, vector<float>* data, int op){
```
nudles: Would it be better to change the signature of Feed to

    Feed(const vector<int>& shape, const vector<float*>& data, vector<int>& aux_data);

Currently, every layer has a `data_` and `aux_data_` fields, including the dummy layer.

src/proto/job.proto
```diff
@@ -647,6 +647,7 @@ enum LayerType {
   kImagePreprocess = 101;
   kPrefetch = 102;
   kRecordInput = 103;
+  kDummyInput = 104;
```
nudles: There is a kDummy, do you need a kDummyInput?

tool/python/examples/user1-cifar10/main.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python
```
nudles: Can we put these two files in a separate branch, since they are used only by Rafiki currently?

include/singa/neuralnet/neuron_layer.h
```diff
@@ -122,13 +122,17 @@ class DropoutLayer : public NeuronLayer {
  */
 class DummyLayer: public NeuronLayer {
  public:
+  void Setup(const std::string str, const vector<Layer*>& srclayers);
   void Setup(const LayerProto& proto, const vector<Layer*>& srclayers) override;
   void ComputeFeature(int flag, const vector<Layer*>& srclayers) override;
   void ComputeGradient(int flag, const vector<Layer*>& srclayers) override;
+  void Feed(int batchsize, vector<float>& data, int is_aux);
```
nudles: I mean the following signature instead of renaming the argument 'op->is_aux'.

Feed(const vector<int>& shape, const vector<float*>& data, const vector<int>& aux_data);

tool/python/examples/train_cifar10.py
```diff
@@ -0,0 +1,117 @@
...
+    print '[Load CIFAR10 dataset]', did
+    dataset_dir_ = singa_root_ + "/examples/cifar10/cifar-10-batches-bin"
+    fname_train_data = dataset_dir_ + "/data_batch_{}.bin".format(did)
+    
+    nb_samples = 10000
+    nb_pixels = 3 * 32 * 32  
+    d = np.fromfile(fname_train_data, dtype=np.uint8)
+    d = d.reshape(nb_samples, nb_pixels + 1) # +1 for label
+    x = d[:, 1:] 
+    x = x - mean_image
```
nudles: It would be better to compute the mean online after loading the data? Currently, we assume that there is a mean file, but it is not known where to get it..
e.g.,
```
    image=...
    label=...
    mean=np.avg(image,...)
    image-=mean
```
There is code for loading cifar10 data (in python format), http://www.cs.toronto.edu/~kriz/cifar.html.
Maybe we can add a usage string telling users where to download the dataset (similar for the mnist dataset).

tool/python/singa/utils/utility.py
```diff
@@ -82,3 +83,9 @@ def setval(proto, **kwargs):
                     fattr.MergeFrom(val)
                 else:
                     setattr(proto, key, val)
+
+def swap32(x):
```
nudles: If this function is only used in parsing MNIST dataset, then we can move it into that file.

tool/python/singa/layer.py
```diff
@@ -219,14 +219,24 @@ def Feed(self, data, nb_channel=1, is_label=0):
...
 
-        data = data.astype(np.float)
-        dataVector = floatVector(datasize)
-        k = 0
-        for i in range(batchsize):
-            for j in range(hdim):
-                dataVector[k] = data[i, j]
-                k += 1
-        self.singalayer.Feed(batchsize, dataVector, is_label)
+        if is_label == 0:
```
nudles: The signature of the init function is not very general.
nb_channel is specific for image data.
can we use the signature the same as the Feed function,
```
Dummy(shape, data, aux_data)
Label(Dummy, labels)
Image(Dummy, shape, images)
```

tool/python/examples/train_cifar10.py
```diff
@@ -28,27 +28,49 @@
...
+    nb_pixels = 3 * 32 * 32  
+    mean = None
+    nb_samples_total = 0
+    for did in range(1,6):
+        fname_train_data = dataset_dir_ + "/data_batch_{}.bin".format(did)
+        d = np.fromfile(fname_train_data, dtype=np.uint8)
+        d = d.reshape(nb_samples, nb_pixels + 1) # +1 for label
+        num = d.shape[0]
+        img = d[:, 1:] 
+        img = img.mean(axis=0)
```
nudles: img's dtype is unint8, then the mean is uint8 (not float).
would there be any effect on the accuracy?

tool/python/examples/train_cifar10.py
```diff
@@ -28,27 +28,49 @@
...
 from singa.layer import *
 from singa.model import *
-from singa.utils.utility import swap32
 
-fname_mean_image = 'tool/python/examples/datasets/cifar10_mean_image'
-mean_image = np.fromfile(fname_mean_image)
+dataset_dir_ = singa_root_ + "/examples/cifar10/cifar-10-batches-bin"
+mean_image = None
+
+def compute_mean_image():
```
nudles: There are two different versions of the cifar10 dataset.
If we use the python version, then the code would be simpler
```
def unpickle(file):
    import cPickle
    fo = open(file, 'rb')
    dict = cPickle.load(fo)
    fo.close()
    return dict

cifar10=unpickle(file)
image=cifar10['data'].astype(dtype=float)
label=np.asarray(cifar10['label'], dtype=int)
mean=np.average(image, axis=0)
image=image-mean[

File: src/neuralnet/layer.cc [hunk 22]
```diff
@@ -33,6 +37,20 @@ namespace singa {
...
 
+void Layer::SetupLayer(Layer* layer, const string str, const vector<Layer*>& srclayers) {
+  LayerProto layer_conf;
+  layer_conf.ParseFromString(str);
+  layer->Setup(layer_conf, srclayers);
+  for (auto param : layer->GetParams())
+      param->InitValues();
+}
+
+Layer* Layer::CreateLayer(const string str) {
```
nudles: can we merge these two functions?

include/singa/neuralnet/neuron_layer.h
```diff
@@ -122,13 +122,17 @@ class DropoutLayer : public NeuronLayer {
  */
 class DummyLayer: public NeuronLayer {
  public:
+  void Setup(const std::string str, const vector<Layer*>& srclayers);
   void Setup(const LayerProto& proto, const vector<Layer*>& srclayers) override;
   void ComputeFeature(int flag, const vector<Layer*>& srclayers) override;
   void ComputeGradient(int flag, const vector<Layer*>& srclayers) override;
+  void Feed(int batchsize, vector<float>& data, vector<int>& aux_data);
+  Layer* ToLayer() { return this;}
```
nudles: Do we need this function?

include/singa/neuralnet/neuralnet.h
```diff
@@ -58,6 +58,13 @@ class NeuralNet {
   static NeuralNet* Create(const NetProto& net_conf, Phase phase,
                            int npartitions);
 
+  /**
+   * Create for python binding, production test mode
+   *
+   */
+  static NeuralNet* CreateNeuralNet(const string str);
+  NeuralNet() {};
```
nudles: Is it used?

PR #137, opened 2016-03-20
SINGA-152 Worker class needs to take greater responsibility of its setting up

undisputed-seraphim:
- Move setup code from Driver::CreateWorkers to Worker::Setup and debug any issues that may arise.


src/worker.cc
```diff
@@ -62,13 +100,11 @@ Worker::~Worker() {
...
-  auto context = Singleton<Context>::Instance();
-  int device = context->device_id(std::this_thread::get_id());
+  //auto context = Singleton<Context>::Instance();
+  //context->SetupDevice(std::this_thread::get_id(), this->device_id_);
+
   LOG(ERROR) << "Worker (group = " << grp_id_ <<", id = " << id_ << ") "
-    << " start on " << (device >= 0 ? "GPU " + std::to_string(device) : "CPU");
-  if (device >= 0)
-    context->ActivateDevice(device);
+    << " start on " << (device_type_ >= 0 ? "GPU " + std::to_string(device_type_) : "CPU");
```
nudles: need to activate the device if it is a GPU?

include/singa/utils/context.h
```diff
@@ -23,6 +23,7 @@
 #define SINGA_UTILS_CONTEXT_H_
```
nudles: I think without device mappings, the Context class cannot support the math blob operations. Therefore, there would be errors from compilation.

nudles:
pls checkout those conf files that are not relevant to this pull request.

include/singa/utils/context.h
undisputed-seraphim: It compiles fine, but execution segfaults in Stub.cc.

PR #145, opened 2016-04-05
SINGA-136 Support cuDNN v4

seaokcs:
cudnn_convolution.cc: Add cudnn Vesrion Macro
Support cuDNN V4


seaokcs:
SINGA-136 Support cuDNN v4

PR #146, opened 2016-04-06
SINGA-130 Data prefetching layer

ug93tad:
Extended StoreInputLayer to support prefetching of data. It maintains a buffer for (key,value) pairs read from the storage layer. In Setup(), it launches a new thread for reading data into the buffer. This thread stores data into the buffer. The ComputeFeature() method waits for thread to finish (join) before parsing it into data_ and aux_ field. Finally, it launches another thread.

In terms of memory consumption, this prefetching use extra (batchsize*recordsize) bytes for the buffer. However, we observed no visible runtime improvement, as I/O time is very small (in order of milliseconds without prefetching, and tens of microsecond with prefetching) compared to CPU time.

Passed cpplint check.


src/neuralnet/input_layer/store.cc
```diff
@@ -38,11 +39,21 @@ void StoreInputLayer::Setup(const LayerProto& conf,
...
+  for (int s : conf.store_conf().shape())
+    shape.push_back(s);
+  data_.Reshape(shape);
+  aux_data_.resize(batchsize_);
+  buf_keys_.resize(batchsize_);
+  buf_vals_.resize(batchsize_);
+
+  // initialize prefetch buffer and start the thread
+  if (conf.store_conf().prefetching())
+    threads_.push_back(thread(&StoreInputLayer::fetch_data, this));
```
nudles: For distributed training, we may have multiple StoreInputLayer.
All these layers are setup on every node to construct the whole neural net.
In that case, we will allocate many prefetch threads and buffers.However, they will not be used on every node.
E.g., input layer A and B will be setup on machine X and Y. But only X will run A and only Y will run B.
The solution is to start the prefetching inside ComputeFeature instead of Setup. X will call A's ComputeFeature, but not call B's ComputeFeature.

include/singa/neuralnet/input_layer.h
```diff
@@ -57,6 +62,8 @@ class StoreInputLayer : virtual public InputLayer {
   int batchsize_ = 1;
   int random_skip_ = 0;
   io::Store* store_ = nullptr;
+  vector<std::string> buf_keys_, buf_vals_;
+  std::deque<std::thread> threads_;  // prefetching thread
```
nudles: may not need a deque? there is only one prefetching thread per store input layer?

PR #149, opened 2016-04-07
Singa-157 Change the priority of cudnn library and install libsingagpu.so

xiezl:
If the cudnn.h is in cuda/include folder, then the --with-cudnn is invalid as cuda/include is looked up first. We need to change the order of these two paths during compilation. On the other hand, we also need to put libsingagpu.so file into a proper folder to make our scripts runnable.

### DIFF
--- a/include/singa/neuralnet/neuron_layer.h
+++ b/include/singa/neuralnet/neuron_layer.h
@@ -478,7 +478,9 @@ class CudnnBMLayer : public BMLayer, public CudnnBase {
   void ComputeFeature(int flag, const vector<Layer*>& srclayers) override;
   void ComputeGradient(int flag, const vector<Layer*>& srclayers) override;
  protected:
+#if CUDNN_MAJOR == 4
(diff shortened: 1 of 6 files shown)


include/singa/neuralnet/neuron_layer.h
```diff
@@ -478,7 +478,9 @@ class CudnnBMLayer : public BMLayer, public CudnnBase {
   void ComputeFeature(int flag, const vector<Layer*>& srclayers) override;
   void ComputeGradient(int flag, const vector<Layer*>& srclayers) override;
  protected:
+#if CUDNN_MAJOR == 4
   cudnnBatchNormMode_t mode_;
+#endif
```
nudles: put the whole cudnnbmlayer into #if  ... #endif.
you can git commit --amend -C HEAD to avoid introducing a new commit.

xiezl: done

PR #151, opened 2016-04-07
SINGA-130 Data prefetching

ug93tad:
Extended StoreInputLayer to support prefetching of data. It maintains a buffer for (key,value) pairs read from the storage layer. In Setup(), it launches a new thread for reading data into the buffer. This thread stores data into the buffer. The ComputeFeature() method waits for thread to finish (join) before parsing it into data_ and aux_ field. Finally, it launches another thread.

In terms of memory consumption, this prefetching use extra (batchsize*recordsize) bytes for the buffer. However, we observe no visible runtime improvement, as I/O time is very small (in order of milliseconds without prefetching, and tens of microsecond with prefetching) compared to CPU time.


src/neuralnet/input_layer/store.cc
```diff
@@ -70,30 +76,44 @@ void StoreInputLayer::ComputeFeature(int flag,
...
+    }
+  }
+}
+
+void StoreInputLayer::ComputeFeature(int flag,
+    const vector<Layer*>& srclayers) {
+  // if prefetching, wait for the thread to finish
+  if (layer_conf_.store_conf().prefetching()) {
+    if (thread_ == nullptr) {
+      buf_keys_.resize(batchsize_);
```
nudles: buf_keys and buf_vals must be allocated no matter prefetching is enabled or not, because they are used by fetch_data().
make test failed...
It works if prefetching's default value is false.
I haven't found the bug..

PR #155, opened 2016-04-29
SINGA-161 DLaaS

aaronwwf:
Wrap SINGA into a Docker image, which can run in a mesos cluster

  Can run in training and testing modes.


nudles:
could you add some instructions on testing it?
thanks.

aaronwwf:
OK

nudles:
I can see a test.sh file, but don't know the arguments that should be passed to it...

PR #159, opened 2016-06-07
SINGA-186

chonho:


chonho:
wrong PR with redundant commits

PR #161, opened 2016-06-07
SINGA-186 Create Python base layer

chonho:
- Add Swig interfaces: singa.i, core_tensor.i
- Add example_tensor.py
- Add generate_singa_wrapper.sh to generate wrapper


src/python/example_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+    if shape == None:
+      self.singa_tensor = singa.Tensor()
+    else:
+      vs = tuple_to_vector(shape)
+      if device == None:
+        self.singa_tensor = singa.Tensor(vs, dtype)
+      else:
+        self.singa_tensor = singa.Tensor(vs, device, dtype)
+
+  def copy_Tensor(self, t):
```
nudles: Does python object have copy function like __copy__?
If not, we can use 
```
def copy(self, t)
```

src/python/example_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+      data_array = singa.intArray_frompointer(self.singa_tensor.intData())
+      dt = np.int32
+    elif idx == kChar:
+      data_array = singa.charArray_frompointer(self.singa_tensor.charData())
+      dt = np.int8
+    elif idx == kDouble:
+      data_array = singa.doubleArray_frompointer(self.singa_tensor.doubleData())
+      dt = np.float64
+
+    data = [data_array[i] for i in range(self.singa_tensor.Size())]
```
nudles: what is the type of data_array?
it may be very slow to convert Tensor data into numpy array in this way.

src/python/example_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+    self.singa_tensor.ToDevice(device)
+
+  def to_host(self):
+    self.singa_tensor.ToHost()
+
+
+  def set_value(self, x):
+    self.singa_tensor.SetValue(x)
+
+  def copy_data_from_hostptr(self, ptr, num):
```
nudles: no need to expose this function python users. It is used by C++ users, who use pointers.
users would call
```
from_array(numpy_array)  
#or
copy_array(numpy_array)
```

src/python/example_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+
+  def copy_data(self, t):
+    self.singa_tensor.CopyData(t.singa_tensor)
+
+  def clone(self):
+    new_t = Tensor(self.tuple_shape, self.device, self.dtype)
+    new_t.singa_tensor = self.singa_tensor.Clone()
+    return new_t
+
+  def matrix_transpose(self):
```
nudles: maybe just use transpose()?

src/python/example_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+    new_t = Tensor(self.tuple_shape, self.device, self.dtype)
+    new_t.singa_tensor = self.singa_tensor.T()
+    return new_t
+
+  # TODO(chonho-02) assign
+  # same as copy_Tensor(t) ???
+  def assign(self, t):
```
nudles: in c++, the copy constructor and assgin operator would actually not copy the data values, instead they just share the input tensor's data pointer (i.e, the blob).
for python, we need to check whether there is a operator=() function and copy constructor. If it has then we should override them like C++.

src/python/example_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+    return call_singa_func(singa.Mul_TT, lhs.singa_tensor, rhs.singa_tensor)
+
+  def __div__(lhs, rhs):
+    return call_singa_func(singa.Div_TT, lhs.singa_tensor, rhs.singa_tensor)
+
+
+  #(TODO) need a method to convert Tensor to numpy array
+  #(TODO) need a method to convert numpy array to Tensor
+
+  def copy_args(self):
```
nudles: why do we need this function? Be careful, it only copies the args, no data is allocated.
it is similar to reset_like().

src/python/example_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+    self.tuple_shape = self.singa_tensor.shape()
+    self.device = self.singa_tensor.device()
+    self.dtype = self.singa_tensor.data_type()
+
+
+''' python functions for global methods in Tensor.h
+'''
+def product(shape):
+  return singa.Product(tuple_to_vector(shape))
```
nudles: no need to call tuple_to_vector, just call python's product function to multiply all elements in the tuple (shape).

src/python/example_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+
+def sizeof(dtype):
+  return singa.SizeOf(dtype)
+
+def call_singa_func(_singa_func, *args):
```
nudles: can we make it private? i.e., users should not call this function.
or need to add comments for it.

src/python/example_tensor.py
```diff
@@ -0,0 +1,325 @@
...
+def __lt__(t1, t2):
+  return singa.Add_T(t1.singa_tensor, t2.singa_tensor)
+'''
+
+def lessthan(t, x):
+  return singa.floatLT_op(t.singa_tensor, x)
+
+
+#---------------------------------------------------------
+# example usage
```
nudles: pls put these example usages into unitest under test/python/.
and rename this file as tensor.py

test/python/test_tensor.py
```diff
@@ -0,0 +1,159 @@
+import sys, os
+import numpy as np
+
```
nudles: For unitest, we can use this one https://docs.python.org/2.7/library/unittest.html.
It is similar to gtest, we use assertEqual to check the execution result the and expected result.

src/python/tensor.py
```diff
@@ -0,0 +1,292 @@
...
+    if device == None:
+      self.singa_tensor = singa.Tensor(vs, dtype)
+    else:
+      self.singa_tensor = singa.Tensor(vs, device, dtype)
+
+    self.tuple_shape = shape
+    self.device = device
+    self.dtype = dtype
+
+  def data(self):
```
nudles: name it as toarray()?

src/python/tensor.py
```diff
@@ -0,0 +1,292 @@
...
+    self.singa_tensor.ToHost()
+
+  def set_value(self, x):
+    self.singa_tensor.SetValue(x)
+
+  def copy_data(self, t):
+    self.singa_tensor.CopyData(t.singa_tensor)
+
+  def clone(self):
+    new_t = Tensor(self.tuple_shape, self.device, self.dtype)
```
nudles: For efficiency, just create an dummy tensor without passing shape, device or dtype.
after call new_t.singa_tensor=..clone..
set shape, device, and dtyep of new_t.

src/python/tensor.py
```diff
@@ -0,0 +1,292 @@
...
+sys.path.append(os.path.join(os.path.dirname(__file__), '../'))
+from core_pb2 import *
+
+
+''' Class and member functions for singa::Tensor
+'''
+class Tensor(object):
+
+  def __init__(self, arg=None, device=None, dtype=kFloat32):
+    ''' arg = Tensor or tuple for shape
```
nudles: may not need to use arg, just use shape.
"Assignment statements in Python do not copy objects, they create bindings between a target and an object"
e.g., Tensor A = B. A is then a reference of B. 

We may add __copy__ for shallow copy which copies the meta data and share the internal data;  and __deepcopy__ for deep copy which calls CopyDataToFrom to copy internal data and meta data.

PR #164, opened 2016-06-11
SINGA-192 Implement optimization algorithms for v1

ijingo:
implement optimization algorithms for Singa v1 including nesterov,
adagrad, rmsprop.
Add unit test cases for these algorithms.
However, only nesterov passed the test case, adagrad and rmsprop need
Sqrt() operation for tensor which has not been implemented yet.

### DIFF
--- a/src/core/tensor/tensor.cc
+++ b/src/core/tensor/tensor.cc
@@ -685,8 +685,7 @@ void Axpy(const SType alpha, const Tensor &in, Tensor *out) {
     }, {in.blob(), out->blob()}, {out->blob()});
   });
 }
-template <>
-void Axpy(const float alpha, const Tensor &in, Tensor *out);
+template void Axpy(const float alpha, const Tensor &in, Tensor *out);
 
 Tensor Mult(const Tensor &A, const Tensor &B) {
   Shape s;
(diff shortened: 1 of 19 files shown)


nudles:
Pls try to run test again to check if any numeric errors in test cases.

src/model/optimizer/adagrad.cc
```diff
@@ -0,0 +1,36 @@
...
+
+void Adagrad::Setup(const OptimizerConf& conf) { delta_ = conf.delta(); }
+
+void Adagrad::Apply(int step, float lr, const string& name, Tensor* grad,
+                    Tensor* value) {
+  if (history_gradient_.find(name) == history_gradient_.end())
+    history_gradient_[name].ResetLike(*value);
+  Tensor& history = history_gradient_[name];
+  history += Square(*grad);
+  (*grad) /= Sqrt(history + delta_);
```
ijingo: I thought we should not cause an side effict on grad after invoking Apply().

src/core/tensor/tensor.cc
```diff
@@ -685,8 +685,7 @@ void Axpy(const SType alpha, const Tensor &in, Tensor *out) {
     }, {in.blob(), out->blob()}, {out->blob()});
   });
 }
-template <>
-void Axpy(const float alpha, const Tensor &in, Tensor *out);
+template void Axpy(const float alpha, const Tensor &in, Tensor *out);
```
ijingo: The explicit instantiation is because the definition of Axpy() function is seperated from its declaration.
We'd better move the definition of template functions into tensor.h
Ref: https://isocpp.org/wiki/faq/templates#templates-defn-vs-decl

PR #165, opened 2016-06-12
SINGA-175 Add memory management APIs and implement a subclass using CNMeM

nudles:
Add CNMem as a submodule in lib/


nudles:
To work on this PR, pls fetch this branch and get the cnmem by
```
git submodule init
git submodule update
```

PR #171, opened 2016-06-14
SINGA-186 Create Python Tensor class

chonho:


src/python/generate_singa_wrapper.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+!/usr/bin/env bash
```
nudles: bash is not available on Windows, need to move these commands to cmake in the future.

src/python/generate_singa_wrapper.sh
```diff
@@ -26,6 +26,8 @@ USR_LOCAL=/home/chonho/local
 
 swig -c++ -python -I../../include singa.i
 
+echo ${SRC_CC[@]}
+
 g++ -fPIC ${SRC_CC[@]} singa_wrap.cxx -shared -o _singa.so \
     -L${USR_LOCAL}/lib -lprotobuf -Wl,-rpath=${USR_LOCAL}/lib \
     -L../../lib -lsinga_core -lsinga_model -lsinga_utils -Wl,-rpath=../../lib \
```
nudles: pls replace the hard-code paths, e.g., SINGA_ROOT and USR_LOCAL.
for -I<LIB_PATH>, the LIB_PATH should use the paths found by cmake.

src/python/generate_singa_wrapper.sh
xiezl: I have already moved the functions in the cmake scripts.
The bash file is only for standalone test.

nudles:
Merged.
TODO:
1. replace hard-code path
2. remove example_xx.py files in python

PR #175, opened 2016-06-16
SINGA-180 - Add Activation layer and Softmax layer

dbxinj:
Fix a bug in cudnn softmax implementation.
Let cudnn_softmax support 1D or 2D tensor as input.

### DIFF
--- a/src/model/layer/cudnn_softmax.cc
+++ b/src/model/layer/cudnn_softmax.cc
@@ -26,30 +26,49 @@ CudnnSoftmax::~CudnnSoftmax() {
   if (desc_ != nullptr) CUDNN_CHECK(cudnnDestroyTensorDescriptor(desc_));
 }
 
-void CudnnSoftmax::InitCudnn(size_t size, DataType dtype) {
+void CudnnSoftmax::Setup(const LayerConf &conf) {
+  Softmax::Setup(conf);
+  SoftmaxConf sft_conf = conf.softmax_conf();
+  std::string algorithm = sft_conf.algorithm();
+  CHECK(algorithm == "accurate" || algorithm == "fast" || algorithm == "log")
+    << "CudnnSoftmax only supports three algorithm preferences: "
+    << "accurate, fast and log.";
+  if (algorithm == "accurate")
+    algorithm_ = CUDNN_SOFTMAX_ACCURATE;
+  else if (algorithm == "fast")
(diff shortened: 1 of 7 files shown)


src/model/layer/cudnn_softmax.cc
```diff
@@ -26,23 +26,48 @@ CudnnSoftmax::~CudnnSoftmax() {
...
+  CHECK_LE(shape.size(), 2u)
+    << "Tensor shape should range from 1 to 2D;"
+    << "otherwise, add flatten layer to transform";
+  if (shape.size() == 1u)
+    CUDNN_CHECK(cudnnSetTensor4dDescriptor( desc_,
+      CUDNN_TENSOR_NCHW, GetCudnnDataType(dtype), 1, shape[0], 1, 1));
+  else
+    CUDNN_CHECK(cudnnSetTensor4dDescriptor( desc_, CUDNN_TENSOR_NCHW,
+      GetCudnnDataType(dtype), shape[0], shape[1], 1, 1));
   has_init_cudnn_ = true;
```
nudles: if you set desc_ as above, then the mode should be fixed to CUDNN_SOFTMAX_MODE_INSTANCE.

test/singa/test_cudnn_softmax.cc
```diff
@@ -33,27 +33,31 @@ TEST(CudnnSoftmax, Setup) {
...
 }
 
 TEST(CudnnSoftmax, Forward) {
-  const float x[] = {1.0f, 2.0f, 0.0f, -2.0f, -3.0f, -1.0};
+  const float x[] = {1.f, 2.f, 0.f, -2.f, -3.f, -1.f};
   size_t n = sizeof(x) / sizeof(float);
+  size_t batch = 2, c = 3;
   singa::CudaGPU cuda(0, 1);
-  singa::Tensor in(singa::Shape{n}, &cuda);
+  singa::Shape shape = {batch, c};
```
nudles: may also try to test 1d input tensor.

dbxinj: I tested 1D first, and then changed to 2D.
Or maybe I can add separate tests for 1D and 2D.

src/model/layer/cudnn_softmax.cc
dbxinj: If we fix mode, do we still need conf param mode?

PR #181, opened 2016-06-22
SINGA-200 - Implement Encoder and Decoder for data pre-processing

dbxinj:
Add base class Encoder and Decoder.
Add inheritance class Image2JPG_Encoder and Image2JPG_Decoder, which use opencv to encode and decode image source files.
One test problem is that encoded images is not exactly the same as the input image because of loss of information.


include/singa/io/encoder.h
```diff
@@ -30,13 +31,15 @@ class Encoder {
   public:
     Encoder() { }
     virtual ~Encoder() { }
+    
+    virtual void Setup(const LayerConf& conf) = 0;
```
nudles: why use LayerConf?

src/io/jpg2proto_encoder.cc
```diff
@@ -16,21 +16,24 @@
...
+
+  std::string JPG2ProtoEncoder::Encode(vector<Tensor>& data) {
     // suppose data[0]: data, data[1]: label
     // data[0] has a shape as {height, width, channel}
     CHECK_EQ(data[0].nDim(), 3u);
     int height = data[0].shape()[0];
     int width = data[0].shape()[1];
     int channel = data[0].shape()[2];
-    Mat mat = Mat(height, width, CV_8UC3);
+    cv::Mat mat = cv::Mat(height, width, CV_8UC3, cv::Scalar(0, 0, 0));
```
nudles: Do you need cv2 or cv?

dbxinj: cv2
#include <opencv2/highgui/highgui.hpp>
#include <opencv2/imgproc/imgproc.hpp>
namespace is still cv

include/singa/io/encoder.h
dbxinj: suppose some parameters are set in LayerConf
currently not sure about what params we will support for users to config

nudles: do not make such assumption.
encoders and decoders should be independent from layer.

src/io/jpg2proto_encoder.cc
nudles: then may need to check the link path. Is it opencv2 or opencv?

src/io/jpg2proto_encoder.cc
```diff
@@ -0,0 +1,70 @@
...
+    CHECK_EQ(data[0].nDim(), 3u);
+    int height = data[0].shape()[0];
+    int width = data[0].shape()[1];
+    int channel = data[0].shape()[2];
+    cv::Mat mat = cv::Mat(height, width, CV_8UC3, cv::Scalar(0, 0, 0));
+    CHECK_EQ(height, mat.rows);
+    CHECK_EQ(width, mat.cols);
+    CHECK_EQ(channel, mat.channels());
+
+    if (data[0].data_type() != kInt)
```
nudles: is this check necessary?

dbxinj: the first check is necessary, we want user to give a 3D input tensor for each raw image.
the rest three checks are not.

PR #185, opened 2016-06-27
SINGA-186 Create Python Tensor class

chonho:
- Update core_tensor.i, tensor.py to take care of the following methods
  . Random operations
  . Axpy
  . Mult
  . Pow
  . AddColumn, SubColumn, MultColumn, DivColumn, and xxxRow, etc.
  . SumColumns, SumRows

- Revised model_layer.i, layer.h to take care of Shape (std::vector)

TODO: from_array() in python

### DIFF
--- a/src/python/tensor.py
+++ b/src/python/tensor.py
@@ -28,6 +28,7 @@ to call singa::Tensor and its methods
 import sys
 import os
 import numpy as np
+import ctypes
 
 sys.path.append(os.path.join(os.path.dirname(__file__),
                              '../../build/lib'))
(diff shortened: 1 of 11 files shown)


src/python/tensor.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, shape=None, device=None, dtype=kFloat32):
             self.device = device
             self.dtype = dtype
 
-    def toarray(self):
+    def to_array(self):
```
nudles: how about improving this using numpy.i like from_array()?

src/python/tensor.py
```diff
@@ -92,6 +92,16 @@ def toarray(self):
         data = np.array(data, dtype=dt).reshape(self.tuple_shape)
         return data
 
+    def from_array(self, np_array, offset=0):
```
nudles: rename it as copy_data_from()? and make from_array a global function (not tensor member function).
similarly for copy_data_to()  --- to_array()

src/python/tensor.py
chonho: Yes right, ARGOUT_ARRAY in numpy.i might be useful. I will try asap

PR #186, opened 2016-06-28
SINGA-211 Add TextFileReader and TextFileWriter for CSV files

XiangruiCAI:
Add TextFileReader and TextFileWriter for reading and writing csv files.
Pass a simple test for these two classes.

Fixed bugs in BinFileReader and BinFileWriter.

### DIFF
--- a/include/singa/io/reader.h
+++ b/include/singa/io/reader.h
@@ -40,7 +40,7 @@ class Reader {
   /// path is the path to the storage, could be a file path, database
   /// connection, or hdfs path.
   /// return true if open successfully, otherwise false.
-  virtual bool Open(const std::string& path, int capacity = 10485760) = 0;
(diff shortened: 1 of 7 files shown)


include/singa/io/reader.h
```diff
@@ -93,6 +95,29 @@ class BinFileReader : public Reader {
   const char kMagicWord[2] = {'s', 'g'};
 };
 
+/// TextFileReader reads tuples from CSV file.
+class TextFileReader : public Reader {
+ public:
+  ~TextFileReader() { Close(); }
+  /// \copydoc Open(const std::string& path)
+  bool Open(const std::string& path, int capacity = 10485760) override;
```
nudles: capacity is not necessary here.
maybe remove the capacity argument.
for BinFileReader, add a 
```
set_buffer_capacity(size_t capacity)
```

PR #192, opened 2016-06-30
SINGA-214 Add LMDBReader and LMDBWriter for LMDB

XiangruiCAI:
Add LMDBReader and LMDBWriter for LMDB. The classes are modified from db_lmdb.cpp and db_lmdb.hpp in Caffe.
Need to swith "USE_LMDB" ON when using these two classes.

Todo: LMDB does not support checking empty and duplicate keys, may support by us later.

### DIFF
--- a/include/singa/io/reader.h
+++ b/include/singa/io/reader.h
@@ -22,6 +22,13 @@
 #include <cstring>
 #include <fstream>
 #include <string>
+#include "singa/singa_config.h"
+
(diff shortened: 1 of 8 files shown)


nudles:
The Reader class may need a SeekToFirst() function for repeating the reading over the dataset multiple times.
Or adding a flag to the read function, e.g., auto_repeat, if true, then seek to first automatically when reaching the file end.

PR #193, opened 2016-07-01
SINGA-213 Implement Encoder and Decoder for CSV

dbxinj:
Implement Encoder and Decoder for CSV file
Write test for it

### DIFF
--- a/src/io/textfile_decoder.cc
+++ b/src/io/textfile_decoder.cc
@@ -16,22 +16,39 @@
  * limitations under the License.
  */
 
-package singa;
+#include "singa/io/decoder.h"
+#include <string>
+#include <sstream>
 
+#define MAXSIZE 4096
 
-message EncoderConf {
-  optional string type = 1 [default = "jpg2proto"];
-  optional string image_dim_order = 2 [default = "HWC"];
+namespace singa {
+
+std::vector<Tensor> TextDecoder::Decode(std::string value) {
+  std::vector<Tensor> output;
+  std::stringstream ss;
+  ss.str(value);
+  int l = 0;
+  if (has_label_ == true) ss >> l;
+  std::string str;
+  float* d = new float[MAXSIZE];
+  size_t size = 0;
+  while(std::getline(ss, str, ',')) {
+    float temp;
+    if (std::stringstream(str) >> temp) {
+      CHECK_LE(size, MAXSIZE-1);
+      d[size++] = temp;
+    }
+  }
+
+  Tensor data(Shape{size}, kFloat32);
+  data.CopyDataFromHostPtr(d, size);
+  output.push_back(data);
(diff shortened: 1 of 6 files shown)


src/io/textfile_decoder.cc
```diff
@@ -0,0 +1,45 @@
...
+
+#include "singa/io/decoder.h"
+#include <string>
+
+namespace singa {
+
+std::vector<Tensor> TextDecoder::Decode(std::string value) {
+  std::vector<Tensor> output;
+  const char* str = value.c_str();
+  const float* d = reinterpret_cast<const float*>(str);
```
nudles: You cannot cast a string to float array...
The string is like "1.2, 2.3, 4.0, ...".
you can use the tokenizer https://github.com/apache/incubator-singa/blob/dev/include/singa/utils/string.h#L77
to extract the float str one by one and convert them into float values.

src/io/textfile_encoder.cc
```diff
@@ -0,0 +1,37 @@
...
+
+std::string TextEncoder::Encode(vector<Tensor>& data) {
+  CHECK_GE(data.size(), 1);
+  size_t size = data[0].Size();
+  const float* value = data[0].data<float>();
+  const char* ptr1 = reinterpret_cast<const char*>(value);
+  std::string d(ptr1, ptr1 + sizeof(float) * size);
+  if (data.size() == 2) {
+    const float label = (const float)data[1].data<int>()[0];
+    const char* ptr2 = reinterpret_cast<const char*>(&label);
```
nudles: do not cast the float* to char*.
the encoded string is not readable. 
csv file is readable.

src/proto/io.proto
```diff
@@ -21,14 +21,26 @@ package singa;
 
 message EncoderConf {
   optional string type = 1 [default = "jpg2proto"];
-  optional string image_dim_order = 2 [default = "HWC"];
+  optional JPG2ProtoEncoderConf jpg2proto_encoder_conf = 2;
```
nudles: no need to use JPG2ProtoEncoderConf. 
the name is too long..

test/singa/test_textfile_decoder.cc
```diff
@@ -0,0 +1,61 @@
...
+  int raw_label = 0;
+  float* raw_data = new float[5];
+  line >> raw_label;
+  size_t size = 0;
+  std::string str;
+  while (std::getline(line, str, ',')) {
+    float temp;
+    if (std::stringstream(str) >> temp)
+      raw_data[size++] = temp;
+  }
```
nudles: you can test the encoder in this way,
1. create a tensor for data and a tensor for label
2. encoder them into string.
3. compare the encoded with the string created by you.

The decoder can be test in 3, 2 (code), 1

PR #194, opened 2016-07-01
SINGA-215 Implement Image Transformation for Image Pre-processing

dbxinj:
Implement some of the image transformations (i.e., crop, resize, horizontal_mirror) for pre-processing stage.

Image resize need to use opencv library.

Currently image transformation only support one sample input each time.

Pass all tests.


include/singa/io/transformer.h
```diff
@@ -0,0 +1,78 @@
...
+
+/// Base transformer class that does data transformations in pre-processing stage.
+class Transformer {
+ public:
+  Transformer() {}
+  virtual ~Transformer() {}
+
+  virtual void Setup(const TransformerConf& conf) {}
+
+  virtual Tensor Transform(Tensor& data) = 0;
```
nudles: rename it to Apply()?

include/singa/io/transformer.h
```diff
@@ -0,0 +1,78 @@
...
+ public:
+  void Setup(const TransformerConf& conf) override {
+    featurewise_center_ = conf.featurewise_center();
+    featurewise_std_norm_ = conf.featurewise_std_norm();
+    resize_height_ = conf.resize_height();
+    resize_width_ = conf.resize_width();
+    rescale_ = conf.rescale();
+    horizontal_mirror_ = conf.horizontal_mirror();
+    image_dim_order_ = conf.image_dim_order();
+    if (conf.crop_shape_size() != 4)
```
nudles: why crop_shape must have 4 values?
pls add some explanations in the comments.

src/io/image_transformer.cc
```diff
@@ -0,0 +1,148 @@
...
+  Tensor ImageTransformer::Transform(Tensor& input) {
+    CHECK_EQ(input.nDim(), 4u);
+    CHECK_EQ(input.data_type(), kFloat32) << "Data type " << input.data_type()
+      << " is invalid for an raw image";
+    
+    /// resize image using opencv resize
+    Tensor temp;
+    size_t samples = input.shape(0);
+#ifdef USE_OPENCV
+    if (resize_height_ > 0 && resize_width_ > 0) {
```
nudles: may split this function into 
resize()
mirror()
crop()
...

include/singa/io/transformer.h
dbxinj: I will add comments in io.proto.
Crop_shape should contain height, width, height_offset and width_offset to specify the position.

include/singa/io/transformer.h
dbxinj: rename transformer class to apply class? why call it apply?

include/singa/io/transformer.h
```diff
@@ -0,0 +1,89 @@
...
+    image_dim_order_ = conf.image_dim_order();
+    
+    /// if crop_shape not contain 2 elements, ignore crop option.
+    if (conf.crop_shape_size() == 2)
+      crop_shape_ = {conf.crop_shape(0), conf.crop_shape(1)};      
+  }
+
+  Tensor Apply(int flag, Tensor& input) override;
+
+  const bool Featurewise_center() const { return featurewise_center_; }
```
nudles: function names could be either  xxx_yyy_zzz() or XxxYyyZzz.
do not use Xxxx_yyyy.

src/io/image_transformer.cc
```diff
@@ -0,0 +1,239 @@
...
+               crop_w_offset = (width - crop_shape_[1]) / 2;
+        temp2 = crop(temp1, crop_shape_[0], crop_shape_[1], 
+                  crop_h_offset, crop_w_offset, image_dim_order_); 
+      }
+    }
+    else temp2 = temp1;
+
+    /// mirror
+    Tensor output;
+    if ((flag == kTrain) && (rand() % 2))
```
nudles: the check for mirror
```
if (horizontal_mirror_ && (flag & kTrain) && (rand() % 2))
```

src/io/image_transformer.cc
```diff
@@ -0,0 +1,239 @@
...
+// #include "singa/singa_config.h"
+
+#ifdef USE_OPENCV
+#include <opencv2/highhui/highhui.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+#endif
+
+namespace singa {
+  Tensor ImageTransformer::Apply(int flag, Tensor& input) {
+    CHECK_EQ(input.nDim(), 4u);
```
nudles: 4u or 3u?
if the input has only one image, then the nDim should be 3u

test/singa/test_image_transformer.cc
```diff
@@ -0,0 +1,119 @@
...
+  img_transformer.Setup(conf);
+  EXPECT_EQ(256, img_transformer.Resize_height());
+  EXPECT_EQ(256, img_transformer.Resize_width());
+  EXPECT_EQ(true, img_transformer.Horizontal_mirror());
+  EXPECT_EQ("NHWC", img_transformer.Image_dim_order());
+  EXPECT_EQ(224u, img_transformer.Crop_shape()[0]);
+  EXPECT_EQ(200u, img_transformer.Crop_shape()[1]);
+}
+
+TEST(ImageTransformer, Apply) {
```
nudles: due to randomness, we cannot test mirror, crop for kEval.
we can do test for kTrain.

src/io/image_transformer.cc
dbxinj: I check it in the mirror function.

src/io/image_transformer.cc
dbxinj: if input one image, then nDim will be like (1, c, h, w) or (1, h, w, c)
later we may extent it to mini-batch transformation

test/singa/test_image_transformer.cc
dbxinj: randomness exists also in crop in kTrain phase.
so, can only test resize in kTrain

src/io/image_transformer.cc
nudles: it is not necessary to crop on the channel dimension.
currently just assume the input image is 3 dim for (c, h, w) and 2d for (h,w).
may add a field like `bool batch_` for users to set later to support batch processing.

PR #204, opened 2016-07-07
SINGA-194 Implementation of the Platform singleton

undisputed-seraphim:
This pull request is a WIP of implementing the Platform class singleton, which is an object that maintains access to the available computing devices on a given machine.


include/singa/core/platform.h
```diff
@@ -0,0 +1,105 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
```
nudles: I put platform declaration in device.h because platform.cc is under src/core/device/.
if you change it, you may also need to change the src/python/swig/core_device.h, and the test/singa/test_platform.cc.
better put it back into device.h for the time being.

include/singa/core/platform.h
```diff
@@ -0,0 +1,105 @@
...
+  CreateCudaGPUs(const size_t num_devices, size_t init_size = 0);
+
+  /// Create a set of CudaGPU Device using given GPU IDs.
+  static const std::vector<std::shared_ptr<Device>>
+  CreateCudaGPUs(const std::vector<int> &devices, size_t init_size = 0);
+
+  /// Create a \p num_devices set of valid OpenCL devices, regardless of platforms.
+  /// If there are fewer valid devices than requested, then this method will return as many as possible.
+  /// If OpenCL is not in use, this method will return an empty array.
+  const std::vector<std::shared_ptr<Device>> CreateOpenclDevices(const size_t num_devices);
```
nudles: all other methods are static.
can you make the CreateOpenclDevices static?

PR #206, opened 2016-07-09
SINGA-222 Fixed bugs in IO

XiangruiCAI:
Fixed a bug in TextFileReader;
Add codes to remove some files generated by LMDB test after testing.
Change URL to travis-ci in README.md.

### DIFF
--- a/test/singa/test_lmdb_rw.cc
+++ b/test/singa/test_lmdb_rw.cc
@@ -132,5 +132,9 @@ TEST(LMDBReader, SeekToFirst) {
   EXPECT_STREQ("This is the fourth test for lmdb io.", value.c_str());
 
   reader.Close();
+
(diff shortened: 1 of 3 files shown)


test/singa/test_lmdb_rw.cc
```diff
@@ -132,5 +132,9 @@ TEST(LMDBReader, SeekToFirst) {
   EXPECT_STREQ("This is the fourth test for lmdb io.", value.c_str());
 
   reader.Close();
+
```
nudles: It seems remove is not available on windows?
may need to delete 136-138.

nudles: It is in std. http://en.cppreference.com/w/cpp/io/c/remove
then just keep it.

PR #208, opened 2016-07-12
SINGA-223 Use Sphinx to create the website.

moazreyad:
The documentation can be generated with Sphinx following the updated  [documentation contribution guidelines.](https://github.com/apache/incubator-singa/blob/a2f3b1e243c71da74b0447a2cce4d52a99845139/doc/develop/contribute-docs.md)

Note that many files are still in markdown (md) format. They must be converted to Sphinx restructuredtext (rst) format later for optimal use of Sphinx.


nudles:
Thanks for the PR.
I think it would be better to convert previous docs one by one instead of all in one PR.
We can add docs for dev branch (v1.0) at first, and then convert docs for v0.3 (after merging dev and master).

PR #211, opened 2016-07-14
A simple naive implementation for parallel training.

ijingo:
This pull request is to get commented.

-  Added a naive Optimizer wrapper class named Updater.
   Updater is used to aggregate gradient from neural net instances, so revise FeedForwardNet class
   a little bit to make it invoke Apply() method provided by Updater instead.
- Revise alexnet.cc in example/cifar10 using a two thread parallel version.
- Fixed a bug in memory.cc


examples/cifar10/alexnet.cc
```diff
@@ -22,16 +22,19 @@
 #include "./cifar10.h"
 #include "singa/model/feed_forward_net.h"
 #include "singa/model/optimizer.h"
+#include "singa/model/updater.h"
 #include "singa/model/initializer.h"
 #include "singa/model/metric.h"
 #include "singa/utils/channel.h"
```
nudles: pls create another file e.g., multi_gpu.cc, and keep the original single gpu training file.

PR #212, opened 2016-07-15
SINGA-215 Implement Image Transformation for Image Pre-processing

dbxinj:
Implement some of the image transformations (i.e., crop, resize, horizontal_mirror) for pre-processing stage.

Image resize need to use opencv library.

Currently image transformation only support one sample input each time.

Add static function resize, crop and mirror for image transformation.

Training phase use random crop and random mirror,
Evaluation phase use central crop and no mirror.

Cannot test phases for mirror because of a random factor.

Fix bugs in test files involving memory allocation and access.

### DIFF
--- a/test/singa/test_image_transformer.cc
+++ b/test/singa/test_image_transformer.cc
@@ -1,0 +1,261 @@
+/************************************************************
+*
+* Licensed to the Apache Software Foundation (ASF) under one
+* or more contributor license agreements.  See the NOTICE file
+* distributed with this work for additional information
+* regarding copyright ownership.  The ASF licenses this file
+* to you under the Apache License, Version 2.0 (the
+* "License"); you may not use this file except in compliance
+* with the License.  You may obtain a copy of the License at
+*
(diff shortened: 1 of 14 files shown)


test/singa/test_image_transformer.cc
```diff
@@ -0,0 +1,261 @@
...
+  in.CopyDataFromHostPtr<float>(x, n);
+  size_t resize_height = 4, resize_width = 5;
+  singa::Tensor out = singa::resize(in, resize_height, resize_width, "HWC");
+  const float* y = out.data<float>();
+
+  cv::Mat mat(height, width, CV_32FC3, cv::Scalar(0, 0, 0));
+  for (size_t i = 0; i < height; i++)
+    for (size_t j = 0; j < width; j++)
+      for (size_t k = 0; k < channel; k++)
+        mat.at<cv::Vec3f>(i, j)[k] = x[i * width * channel + j * channel + k];
```
nudles: TODO better check the datatype of the tensors, uint8 or float.
Tensor::AsType could be helpful for data type change.

PR #216, opened 2016-07-19
Add parallel training on a single machine for singa v1.0

ijingo:
This commit implements a updater class for parallel training
in local-cpu and local-gpu mode. (The mode specification described
in https://issues.apache.org/jira/browse/SINGA-226)

Updater class is a wrapper of Optimizer class. It controls the
communication pattern among workers. When initializing Updater,
the user needs to provide the total number of workers and where
the Updater does aggregation and computation.

File changed descibed as follows:
* Put a new folder named cifar-parallel under example, containing
  the single machine multi-gpu parallel training example.
* Replace Optimizer pointer in feed_forward_net class with Updater
  pointer, since Updater is a wrapper of Optimizer. So the compile()
  method is changed accordingly.
* Add a helper function TrainThread() method in feed_forward_net to
  launch a training thread.
* Adapt alexnet.cc original cifar10 example to support the new compile()
* Fixed a bug in memory.cc which happens during initialzing cnmem
  memory pool


CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 2.6)
 
 PROJECT(singa)
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -lpthread")
```
ijingo: Leave out this change pls...

nudles:
I just tested the time for running with 2gpu and 1gpu (140epochs in total):
2gpus 14 min
1gpu 8 min
Need to figure out the reasons.

ijingo:
It seems copy gradient to cpu and do sum operation on cpu cost times.
Following is time cost in each part when train on one batch:

Forward time: 0.000988s  (on gpu)
Backward time: 0.001189s (on gpu)
Update time: 0.014970s (on cpu)

In update time:
Copy gradient to cpu 0.004150s
add gradient 0.001950s
average gradient 0.000852s
copy parameter to cpu 0.000891s
optimization algorithm on cpu 0.006571s
copy updated parameter back to gpu 0.000556s

src/model/updater/local_updater.cc
```diff
@@ -0,0 +1,74 @@
...
+  opt_->Register(name, specs);
+  param_buffer_[name];
+  param_buffer_[name].ToDevice(dev_);
+  sum_[name];
+  sum_[name].ToDevice(dev_);
+  for (int i = 0; i < total_num_; ++i) {
+    grad_buffer_[std::make_pair(i, name)];
+    grad_buffer_[std::make_pair(i, name)].ToDevice(dev_);
+  }
+  dev_index_[name] = 0;
```
nudles: add
```
to_updater_finished_[name] = 0;
```
could avoid the dead-clock.

PR #219, opened 2016-07-19
173 OpenCL device support and implementation

undisputed-seraphim:
This pull request merges in OpenCL device support.

- OpenclDevice (subclass of Device) is implemented.
-   OpenCL kernels and unit tests written. Note that kernels for random value generation is in `distribution.cl`.
- A workaround to make Travis-CI tests pass has also been implemented. Travis-CI currently does not support OpenCL well (not without a lot of strange workarounds) so instead I added a switch in CMake to skip OpenCL unit tests on Travis.

Some shortcomings, notes, and future work:

 -  Unit tests for OpenCL will randomly segfault if the unit tests are run individually (with gtest_filter). I suspect this is due to the shared_ptr being used in the constructor. They run fine if you run all of them at once.
 -  Current implementation of OpenclDevice only chooses the first platform and first device encountered. This is expected to be fixed in SINGA-192 Platform Singleton, where a new class will handle multiple OpenCL platforms and devices in the same computer and choose one according to user configuration.
 -  When the CUDA device and the OpenCL device are the same, I haven't tested that.

If you want to test OpenCL on Intel CPUs or GPUs, you can download the required files [here.](https://software.intel.com/en-us/articles/opencl-drivers) Intel and Nvidia's support for OpenCL is only up to version 1.2, so that is what we'll work with.

I believe AMD has support for 2.0 and beyond, but I don't have any AMD C/GPUs to test this with so I don't know.


include/singa/utils/context.h
```diff
@@ -47,7 +47,7 @@ CHECK_EQ(error, cudaSuccess) << " " << cudaGetErrorString(error); \
 #include <cudnn.h>
```
nudles: This file is not used and will be removed soon.
If your code has dependency on it, please resolve it.

include/singa/core/platform.h
```diff
@@ -0,0 +1,105 @@
+/**
```
nudles: can you move the platform class back to device.h?
because other code uses include "device.h" to get the Platform class.

include/singa/utils/context.h
undisputed-seraphim: My code does not modify anything from cudnn, so it is okay to ignore.

nudles:
```
(py2)wangwei@slave2:~/singav1/incubator-singa/test/python$ python test_tensor.py
  File "test_tensor.py", line 27, in <module>
    from singa import tensor
  File "../../build/python/singa/tensor.py", line 25, in <module>
    from . import singa_wrap as singa
  File "../../build/python/singa/singa_wrap.py", line 28, in <module>
    _singa_wrap = swig_import_helper()
  File "../../build/python/singa/singa_wrap.py", line 24, in swig_import_helper
    _mod = imp.load_module('_singa_wrap', fp, pathname, description)
ImportError: ../../build/python/singa/_singa_wrap.so: undefined symbol: _ZN5singa8PlatformC1Ev
```

PR #224, opened 2016-07-28
Singa-228 Add Cpp Version of Convolution and Pooling layer

XiangruiCAI:
Implement cpp version of convolution layer and pooling layer.
Using some functions in previous version of Singa.
Pooling layer only supports max and average pooling(no stochastic).
Passed tests.

### DIFF
--- a/src/model/layer/convolution.cc
+++ b/src/model/layer/convolution.cc
@@ -24,7 +24,7 @@ namespace singa {
 using std::vector;
 
 RegisterLayerClass(Convolution);
-void Convolution::Setup(const Shape& in_sample, const LayerConf &conf) {
+void Convolution::Setup(const Shape &in_sample, const LayerConf &conf) {
   Layer::Setup(in_sample, conf);
   ConvolutionConf conv_conf = conf.convolution_conf();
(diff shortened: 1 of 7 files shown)


src/model/layer/convolution.cc
```diff
@@ -97,24 +97,136 @@ void Convolution::Setup(const Shape& in_sample, const LayerConf &conf) {
...
+  size_t imagesize = input.Size() / batchsize;
+  DataType dtype = input.data_type();
+  auto dev = input.device();
+  Shape shape{batchsize, num_filters_, conv_height_, conv_width_};
+  Tensor output(shape, dev, dtype);
   Tensor col_data(Shape{col_height_, col_width_});
-  Tensor col_grad(Shape{col_height_, col_width_});
+  float *data_col = new float[col_height_ * col_width_];
+  auto in_data = input.data<float>();
+  for (size_t b = 0; b < batchsize; b++) {
```
nudles: according to http://arxiv.org/abs/1606.04487, if we do im2col for all images inside one mini-batch and then do gemm, the performance would be much better.
may apply such optimization in the future.

PR #226, opened 2016-07-29
SINGA-227 Add Split and Merge Layer and add ResNet Implementation

dbxinj:
ResNet implementation is based on the torch re-implementation of resnet-20 for cifar-10 dataset.

Make changes to the FeedForwardNet Functions (ie. add, forward, backward).


nudles:
pls refer to the examples/cifar10/vgg for the usage of batch normalization layer.

PR #230, opened 2016-08-04
SINGA-233 New communication interface for SINGA

cac2003:
In this ticket, we design and implement a new communication interface for SINGA. Specifically, each communication entity (physical node) is abstracted into an EndPoint, and users can directly call its send/recv method to accomplish data transfer.


nudles:
pls update the commit msg (title and content).

test/CMakeLists.txt
```diff
@@ -9,9 +9,14 @@ IF(NOT USE_OPENCL)
     LIST(REMOVE_ITEM singa_test_source "singa/test_opencl.cc")
 ENDIF()
 
-ADD_EXECUTABLE(test_singa "gtest/gtest_main.cc" ${singa_test_source})
-ADD_DEPENDENCIES(test_singa singa_core singa_utils)
-MESSAGE(STATUS "link libs" ${singa_linker_libs})
-TARGET_LINK_LIBRARIES(test_singa gtest singa_core singa_utils singa_model
+#ADD_EXECUTABLE(test_singa "gtest/gtest_main.cc" ${singa_test_source})
```
nudles: pls uncomment these lines as they are necessary for compiling other tests.

test/singa/test_ep.cc
```diff
@@ -0,0 +1,92 @@
+#include "singa/io/network/endpoint.h"
```
nudles: pls add the license header to each new file.

include/singa/io/network/endpoint.h
```diff
@@ -0,0 +1,139 @@
+/************************************************************
+*
```
nudles: would it be better to include all header files of network into a single include/singa/io/network.h file, and put the .cc files into src/io/network/ folder?

include/singa/io/network/endpoint.h
```diff
@@ -0,0 +1,139 @@
...
+#define MAX_RETRY_CNT 3
+
+#define EP_TIMEOUT 5.
+
+class NetworkThread;
+class EndPointFactory;
+
+class EndPoint {
+    private:
+        std::queue<Message*> send_;
```
nudles: we follow google coding style with 2space indent and max line width=80.
I can re-format the code.

include/singa/io/network/endpoint.h
cac2003: fine with me. I will do that...

PR #232, opened 2016-08-08
SINGA-232 Alexnet on Imagenet

XiangruiCAI:
Implement Alexnet on Imagenet.
1. The model is following the imagenet paper.
2. The data is created offline, including multiple training files,
  one test file and one mean file. All of them are in binary format.
  This part is implemented via writer, encoder in SINGA.
3. Loading data in multiple threads. This part needs reader,
  decoder and transformer in SINGA.
4. This example need OpenCV support.
5. snapshot, jpgencoder, timer, binfile_reader are slightly modified.


examples/imagenet/alexnet.cc
```diff
@@ -0,0 +1,410 @@
...
+  pos = singa::ArgPos(argc, argv, "-pfreq");
+  size_t pfreq = 100;
+  if (pos != -1) pfreq = atoi(argv[pos + 1]);
+
+  pos = singa::ArgPos(argc, argv, "-nthreads");
+  int nthreads = 12;
+  if (pos != -1) nthreads = atoi(argv[pos + 1]);
+
+  LOG(INFO) << "Start training";
+  singa::Train(nEpoch, lr, batchsize, train_file_size, bin_folder,
```
nudles: it would be better to remove num_train_images and num_test_images and filesize as necessary arguments.
as long as the training and test data are stored in train/ and test/ folder, we can get all of them without the above three arguments.

nudles: may update it in the future.

examples/imagenet/alexnet.cc
```diff
@@ -0,0 +1,410 @@
...
+          GenConvConf("conv4", 384, 3, 1, 1, 0.01, 1.0));
+  net.Add(new CudnnActivation(), GenReLUConf("relu4"));
+  net.Add(new CudnnConvolution(),
+          GenConvConf("conv5", 256, 3, 1, 1, 0.01, 1.0));
+  net.Add(new CudnnActivation(), GenReLUConf("relu5"));
+  net.Add(new CudnnPooling(), GenPoolingConf("pool5", true, 3, 2, 0));
+  net.Add(new Flatten(), GenFlattenConf("flat"));
+  net.Add(new Dense(), GenDenseConf("ip6", 4096, 0.005, 1, 1.0));
+  net.Add(new CudnnActivation(), GenReLUConf("relu6"));
+  net.Add(new CudnnDropout(), GenDropoutConf("drop6", 0.5));
```
nudles: cudnndropout is not available for version<5. need to add a macro check.
```
#if CUDNN_VERSION_MAJOR >=5
...
#endif
```

or replace CudnnDropout with Dropout.

PR #238, opened 2016-08-11
SINGA-230 OpenCL Convolution and Pooling Layers

undisputed-seraphim:
This is the pull request for Issue 230 in JIRA, for the creation of OpenCL Convolution and Pooling layers as well as the respective unit tests.


src/model/layer/opencl_convolution.cc
```diff
@@ -0,0 +1,256 @@
...
+  auto batchsize = input.shape(0);
+  auto imagesize = input.Size() / batchsize;
+  auto data_type = input.data_type();
+  auto device = input.device();
+  
+  Shape shape{batchsize, num_filters_, conv_height_, conv_width_};
+  Tensor output(shape, device, data_type);
+  Tensor col_data(Shape{col_height_, col_width_}, device, data_type);
+  
+  for (size_t b = 0; b < batchsize; b++) {
```
nudles: opencl process all images in one batch in parallel.
never mind. let's do the optimization later.

PR #253, opened 2016-09-07
Fix some typo and bug

RUAN0007:
Despite for some typos, it seems tensor does not support the operation like float * tensor. But tensor * float works.


nudles:
could you pls create a jira ticket or update the title to make it more clear, e.g. 'bug from using float * tensor in loss.py'?
thanks.

nudles:
Pls also update the commit title using the title of https://issues.apache.org/jira/browse/SINGA-245 to make it similar as previous commits.

PR #254, opened 2016-09-08
 SINGA-236 memory pool (update from an old branch)

liyuchenmike:
Implemented the following features：

1.A memory pool facility to management Block data allocated by the memory pool.
2.Add relevant test cases

### DIFF
--- a/include/singa/core/memory.h
+++ b/include/singa/core/memory.h
@@ -23,6 +23,7 @@
 #include <atomic>
 #include "singa/proto/core.pb.h"
 #include "singa/singa_config.h"
+#include "singa/core/common.h"
 
 #ifdef USE_CUDA
 #include "cnmem.h"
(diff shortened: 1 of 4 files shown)


include/singa/core/memory.h
```diff
@@ -50,6 +51,57 @@ class DeviceMemPool {
 //  size_t init_size_ = 0, max_size_ = 0;
 };
 
+class CppMemPool {
+	public:
+		// initial pool size (MB), and the size of each memory uint in the memory pool (KB)
+		CppMemPool(size_t init_size_mb = 256, size_t uint_size_kb = 1);
```
nudles: Are all memory units of the same size?
Can we do it in this way:
1. create a large memory pool P0 at the beginning
2. when malloc is called, return a block of the asked size and move the offset of the start pointer of the pool Pi
3. if not enough space for malloc a new block, then create a new block Pi+1.

nudles: optimizations could be added later, e.g.
OptimizeMemPool():  re-compute the total size blocks in use, and move them into contiguous memory space.

include/singa/core/memory.h
```diff
@@ -50,6 +51,57 @@ class DeviceMemPool {
...
+		// return a new pool based on the current pool
+		// once returned, the old pool will be invalid
+		// re-initial with pool size (MB), and set the size of each memory uint in the memory pool (KB)
+		void RsetMemPool(size_t init_size_mb = 256, size_t uint_size_kb = 1);
+
+		// create the memory requested, if size is larger than memUintSize, malloc from system call
+		// is_ptr_null indicate whether the pointer is null and if so we will initialize it in the malloc function,
+		// otherwise we will use the ptr directly and access its data and functions.
+		// after the malloc, the data pointer of the block will be changed and the orginal data pointer will be lost.
+		void Malloc(Block** ptr, const size_t size, bool is_ptr_null = true);
```
nudles: would it be simpler if the API is 
Block* Malloc(const size_t size)

liyuchenmike:
Updated the memory pool structure.
The APIs are simpler and no fixed allocation is required.
Ready to merge.

PR #255, opened 2016-09-14
SINGA-236 memory pool

liyuchenmike:
Update the memory pool.

1. Simplify the APIs.
2. No fixed allocation is required.

### DIFF
--- a/src/core/memory/memory.cc
+++ b/src/core/memory/memory.cc
@@ -20,9 +20,123 @@
 #include "singa/utils/logging.h"
 #include "singa/proto/core.pb.h"
 #include <iostream>
+/*
+int get_pos(size_t size) {
+	int result = 0;
+	while(size > 1) {
+		result++;
+		size = size/2;
+	}
+	return result;
+}
+*/
+namespace singa {
+
+CppMemPool::CppMemPool()	{
+	memPoolSize = 0;
+	freeSize = 0;
+	ppAllocUints = (struct _Uint**)malloc(64*sizeof(struct _Uint*));
+	ppFreeUints = (struct _Uint**)malloc(64*sizeof(struct _Uint*));
+	for(int i = 0; i < 64; i++) {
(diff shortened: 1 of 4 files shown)


src/core/memory/memory.cc
```diff
@@ -20,9 +20,123 @@
...
+	if(pFreeUint == NULL) { // if no available free blocks
+		memPoolSize += blkSize;
+		pCurUint = (struct _Uint*)malloc(blkSize);
+		pCurUint->pPrev = NULL;
+		pCurUint->pNext = pAllocUint; 
+		if(pAllocUint != NULL) {
+			pAllocUint->pPrev = pCurUint;
+		}
+		pAllocUint = pCurUint;
+		pAllocBlk = new Block((char*)(pCurUint) + sizeof(struct _Uint), size);
```
nudles: size of the allocated block + sizeof (struct _Uint) == size?

src/core/memory/memory.cc
```diff
@@ -20,9 +20,123 @@
...
+		if(pFreeUint != NULL) {
+			pFreeUint->pPrev = NULL;
+		}
+		
+		pCurUint->pNext = pAllocUint;
+		if(pAllocUint != NULL) {
+			pAllocUint->pPrev = pCurUint;
+		}
+		pAllocUint = pCurUint;
+		pAllocBlk = pCurUint->pBlk;
```
nudles: pAllocBlk->size() could be larger than size?

src/core/memory/memory.cc
liyuchenmike: If the requested size is 1024, then we allocated 1024+16, 16 is used for metadata of the mempool.

src/core/memory/memory.cc
liyuchenmike: yes, I will reset the size. Thanks for identifying the bug.

PR #257, opened 2016-09-18
SINGA-247 Add windows support for singa

xiezl:
Add MSVC codes in cmake files.
Add DISABLE_WARNINGS into configure files.
Fix incompatible codes and add some definitions.


src/core/memory/memory.cc
```diff
@@ -15,13 +15,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#ifndef DISABLE_WARNINGS
 
+#ifdef USE_CUDA
```
nudles: USE_CUDA is defined in the singa_config.h file, which is included in memory.h.
should put memory.h before the macro

src/io/jpg_decoder.cc
```diff
@@ -15,11 +15,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#ifdef USE_OPENCV
```
nudles: the same as USE_CUDA. 
need to include "singa_config.h"

src/model/layer/pooling.cc
```diff
@@ -85,15 +85,15 @@ const Tensor Pooling::Forward(int flag, const Tensor& input) {
     Tensor mask;
     mask.ResetLike(output);
     float* maskptr = new float[mask.Size()];
-    ForwardMaxPooling(inptr, batchsize, channels_, height_, width_, kernel_h_,
-                      kernel_w_, pad_h_, pad_w_, stride_h_, stride_w_, outptr,
+    ForwardMaxPooling(inptr, (int)batchsize, (int)channels_, (int)height_, (int)width_, (int)kernel_h_,
```
nudles: it is not a good solution to do type cast in this way.
maybe we can add this in singa_config.h
```
#if wind32
typedef size_t unsigned int
#endif
```

src/utils/logging.cc
```diff
@@ -23,7 +23,11 @@
 
 #include <stdlib.h>
 #include <sys/types.h>
+#ifdef _MSC_VER
+#include <io.h>
```
nudles: logging.cc should not have dependency on io.h?

test/CMakeLists.txt
```diff
@@ -37,9 +37,9 @@ ENDIF()
 
 
 ADD_EXECUTABLE(test_singa "gtest/gtest_main.cc" ${singa_test_source})
-ADD_DEPENDENCIES(test_singa singa)
+ADD_DEPENDENCIES(test_singa singa_static)
```
nudles: have you tried libsinga.so instead of singa_static?

doc/en/docs/installation.md
```diff
@@ -214,8 +214,46 @@ To build SINGA with OpenCL support, you need to pass the flag during cmake:
...
+    $ md build && cd build
+    $ cmake -G "Visual Studio 14 Win64" -DUSE_CUDA=OFF -DUSE_PYTHON=OFF ..
+
+If you get error outputs like "Could NOT find xxxxx" indicating a dependent
+library missing, please configure your library file and include path for cmake or the system.
+For example, you get an error "Could NOT find CBLAS" and suppose you installed
+openblas header files at "d:\include" and openblas library at "d:\lib". You should run the
+following command to specify your cblas parameters in cmake:
+
+    $ cmake -G "Visual Studio 14" -DUSE_CUDA=OFF -DUSE_PYTHON=OFF
```
nudles: can we export CMAKE_INCLUDE_PATH and CMAKE_LIBRARY_PATH as in linux?

test/CMakeLists.txt
xiezl: There are two libraries here, singa_static and singa. In linux, there will be libsinga_static.a and libsinga.so.

src/utils/logging.cc
xiezl: io.h is used to replace unistd.h in windows. There is no header file named unistd.h in msvc.

src/core/memory/memory.cc
xiezl: Those macros are for avoiding warnings. I thought the macros will be changed once the code in the source file is completed.

doc/en/docs/installation.md
xiezl: I tried but I failed in windows.

src/core/memory/memory.cc
xiezl: In addition, the singa_config.h file is already included in core/memory.h.

nudles: I mean the macro should be after include "singa/core/memory.h" or after include "singa/singa_config.h"

src/utils/logging.cc
nudles: Is <io.h> a standard header file in windows?
If not, which lib should be installed to get io.h?

test/CMakeLists.txt
nudles: Is it ok to use singa here?

src/core/memory/memory.cc
nudles: but the macro check is before include "singa/core/memory.h"..
I think we should include memory.h before macro.

doc/en/docs/dependencies.md
```diff
@@ -0,0 +1,79 @@
...
+
+Tested on version 2.6.1:
+
+Step 1: Download and decompress the source code.
+
+Step 2: Open "protobuf.sln" file under "vsprojects" folder. You will get a conversion
+dialog and please finish it by the prompts. Compile all the projects in the solution after proper
+configuration. Especially "libprotobuf", "libprotobuf-lite", "libprotoc" and
+"protoc" projects.
+
```
nudles: Error from compiling protobuf 2.6.1 from source in vs 2015.
```
Severity    Code    Description    Project    File    Line    Suppression State
Error    C2338    <hash_set> is deprecated and will be REMOVED. Please use <unordered_set>. You can define _SILENCE_STDEXT_HASH_DEPRECATION_WARNINGS to acknowledge that you have received this warning.    libprotobuf    c:\program files\microsoft visual studio 14.0\vc\include\hash_set    17    
```
It seems the bug is fixed after releasing 2.6.1.

xiezl: Solution: Right click on libprotobuf project -> Properties -> C/C++ -> Preprocessor -> Add "_SILENCE_STDEXT_HASH_DEPRECATION_WARNINGS" to Preprocessor Definitions.

src/core/memory/memory.cc
```diff
@@ -15,13 +15,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#ifndef DISABLE_WARNINGS
```
nudles: should put the macro after include "singa/singa_config.h"?

src/model/layer/pooling.cc
nudles: Is it warning or error if no cast for int, size_t, unsigned int?

PR #258, opened 2016-09-21
SINGA-230 OpenCL Pooling and Convolution layers

undisputed-seraphim:
Sorry it took so long :/ Debugging took a lot of time.

 - OpenCL Pooling and Convolution layers implemented and kernel files added.
 - Unit tests written to support above files
 - There was a rather severe implementation bug in OpenCL GEMV and GEMM, which has been fixed.


nudles:
Could you add license header to the newly added files?

include/singa/core/device.h
```diff
@@ -214,9 +214,9 @@ class OpenclDevice : public singa::Device {
 // Overridden, inherited methods
   void SetRandSeed(unsigned seed) override;
 
-  void CopyDataToFrom(Block* dst, Block* src, size_t nBytes,
+  virtual void CopyDataToFrom(Block* dst, Block* src, size_t nBytes,
```
nudles: please remove 'virtual', which is not need.

PR #260, opened 2016-09-29
SINGA-253 Net converter for caffe model

XiangruiCAI:
Convert caffe model into singa model.
It is a very basic implementation. Now it can convert feed forward net and supports pysinga.
Implementation method:
1. read proto file of caffe model using caffe.proto, and serialize it to string.
2. parse the string use singa model.proto and get the layer config.
3. setup each layer and add it to a feed forward net.


python/singa/layer.py
```diff
@@ -78,12 +78,54 @@ class Layer(object):
...
-        self.param_specs = []
+    def __init__(self, name, conf = None, **kwargs):
+        if conf == None:
+            self.layer = None  # layer converted by swig
+            self.name = name  # TODO(wangwei) duplicate with self.conf.name
+            self.conf = model_pb2.LayerConf()
+            self.conf.name = name
+            self.param_specs = []
+        else:
+            self.layer = None
```
nudles: move these into a class method, e.g. create_caffe_layer()?

python/singa/layer.py
```diff
@@ -78,12 +78,54 @@ class Layer(object):
...
+            self.name = conf.name
+            self.conf = conf
+            # TODO(Xiangrui) need to fill in param_specs
+            self.param_specs = []
+            if conf.type == 'Convolution' or conf.type == 4:
+                _check_engine(engine, ['cudnn', 'singacpp'])
```
nudles: it seems all the if-else are for engine check?
we may add a function covert_engine(caffe_engine) -> signa_engine, and create each type of layer as 
```
eng = convert_engine(caffe_engine)
layer_type = convert_layer_layer(conf.type)
self.layer = __create_layer(engine, layer_type)
```
Dense layer needs special processing.

tool/caffe_model/caffe_parse/caffe.proto
```diff
@@ -0,0 +1,1399 @@
+syntax = "proto2";
```
nudles: move this into src/proto/caffe.proto?
better add the commit of caffe from which the caffe.proto is copied.

tool/caffe_model/cifar10_full_train_test.prototxt
```diff
@@ -0,0 +1,220 @@
+name: "CIFAR10_full"
```
nudles: move this into examples/cifar10/caffe
and add another file, e.g. caffe.py for training using the prototxt files.

tool/caffe_model/Makefile
```diff
@@ -0,0 +1,17 @@
+# find protoc
```
nudles: let the makefile in python to generate the pb2 file for caffe proto.

tool/caffe_model/convert.py
```diff
@@ -0,0 +1,110 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
nudles: move it into python/ as caffe_converter.py.
users can then use it to create caffe nets and then start training by providing data.

PR #262, opened 2016-10-02
SINGA-254 Implement Adam for V1

nudles:
Implemented Adam for pysinga.
Added an argument to control the applying of regularization and constraint
for parameters, which is necessary for BN layers as the mean and std
should not be updated by updaters. The effect is not obvious for other
udpaters, but is very significant for Adam.

Tested Adam for alexnet on cifar10, and the accuracy was 0.8 (SGD is
0.82); It was also not as good as SGD on VGG net (0.92); May need to
tune the learning rate.

To test this PR, 
1. please replace the optimizer from SGD to Adam in examples/cifar10/train.py
2. pass a constant learning rate (e.g 0.01) to apply_with_lr, instead of the learning rate generation function
3. add the iteration ID as the last argument of apply_with_lr.

### DIFF
--- a/python/singa/optimizer.py
+++ b/python/singa/optimizer.py
@@ -32,6 +32,7 @@ Example usage::
   sgd.apply_with_lr(2, 0.03, g, p, 'param')  # use lr=0.03 for epoch 2
 '''
 
+import math
 from . import singa_wrap as singa
 import tensor
 from proto import model_pb2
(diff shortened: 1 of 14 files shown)


python/singa/optimizer.py
```diff
@@ -251,47 +287,70 @@ class AdaGrad(Optimizer):
...
+        if name not in self.m or name not in self.v:
+            self.m[name] = tensor.Tensor(grad.shape, grad.device, grad.dtype)
+            self.m[name].set_value(0)
+            self.v[name] = tensor.Tensor(grad.shape, grad.device, grad.dtype)
+            self.v[name].set_value(0)
+
+        self.m[name] *= self.beta_1
+        tensor.axpy(1 - self.beta_1, grad, self.m[name])
+        self.v[name] *= self.beta_2
+        tensor.axpy(1 - self.beta_1, tensor.square(grad), self.v[name])
```
ijingo: typo: should be "tensor.axpy(1 - self.beta_2, tensor.square(grad), self.v[name])"

PR #269, opened 2016-10-17
SINGA-253 Net converter for caffe model

XiangruiCAI:
Convert caffe model into singa model.
It is a very basic implementation. Now it can convert feed forward net and supports pysinga.
Implementation method:
1. read proto file of caffe model using caffe.proto, and serialize it to string.
2. parse the string use singa model.proto and get the layer config.
3. setup each layer and add it to a feed forward net.

Update and example:
1. put conversion funcions into layer.py and converter.py;
2. add 'caffe' option in cifar example which converts alexnet from
caffe to singa.

### DIFF
--- a/python/singa/converter.py
+++ b/python/singa/converter.py
@@ -1,0 +1,155 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# =============================================================================
+
+from google.protobuf import text_format
+from singa import layer
+from singa import metric
+from singa import loss
+from singa import net as ffnet
+from .proto import model_pb2
+from .proto import caffe_pb2
+
+
+class CaffeConverter:
+
+    def __init__(self, net_proto, solver_proto = None, input_sample_shape = None):
+        self.caffe_net_path = net_proto
+        self.caffe_solver_path = solver_proto
+        self.input_sample_shape = input_sample_shape
+
+    def read_net_proto(self):
+        net_config = caffe_pb2.NetParameter()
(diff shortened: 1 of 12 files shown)


python/singa/converter.py
```diff
@@ -0,0 +1,146 @@
...
+from google.protobuf import text_format
+from singa import layer
+from singa import metric
+from singa import loss
+from singa import net as ffnet
+from .proto import model_pb2
+from .proto import caffe_pb2
+
+
+class caffeConverter:
```
nudles: caffeConverter-->CaffeConverter

python/singa/converter.py
```diff
@@ -0,0 +1,146 @@
...
+        return parser_object
+
+    def convert_engine(self, layer_conf, caffe_solver):
+        '''
+        Convert caffe engine into singa engine
+        return:
+            a singa engine string
+        '''
+        caffe_engine = ''
+        solver_mode = caffe_solver.solver_mode
```
nudles: may pass the solver_model directly, i.e. replace caffe _solver->solver_mode

python/singa/converter.py
```diff
@@ -0,0 +1,146 @@
...
+        if len(caffe_net.layer):
+            layer_confs = caffe_net.layer
+        elif len(caffe_net.layers):
+            layer_confs = caffe_net.layers
+        else:
+            raise Exception('Invalid proto file!')
+
+        net = ffnet.FeedForwardNet()
+        for i in range(len(layer_confs)):
+            if layer_confs[i].type == 'Data' or layer_confs[i].type == 5:
```
nudles: move data, loss and accuracy out of FeedForwardNet?
Net only consists of layers.

add another function for parsing solver to get loss and metric instances?

python/singa/converter.py
```diff
@@ -0,0 +1,146 @@
...
+        '''
+        Create singa net based on caffe proto files.
+            net_proto: caffe prototxt that describes net
+            solver_proto: caffe prototxt that describe solver
+            input_sample_shape: shape of input data tensor
+        return:
+            a FeedForwardNet object
+        '''
+        caffe_net = self.read_net_proto()
+        caffe_solver = self.read_solver_proto()
```
nudles: Maybe better make solver optional. 
For the prediction phase, we only need the net proto.
If it is not available, then just set solver_model to cpu.

python/singa/converter.py
```diff
@@ -0,0 +1,146 @@
...
+from singa import metric
+from singa import loss
+from singa import net as ffnet
+from .proto import model_pb2
+from .proto import caffe_pb2
+
+
+class caffeConverter:
+
+    def __init__(self, net_proto, solver_proto, input_sample_shape):
```
nudles: solver_proto=None
input_sample_shape=None
If the net proto has the input shape, then we don't need to pass in it.

XiangruiCAI:
code updated, thanks!

PR #277, opened 2016-11-16
[SINGA-268] Add IPython notebooks to the documentation

moazreyad:


doc/notebook/rnn.ipynb
```diff
@@ -0,0 +1,151 @@
+{
```
nudles: RNN training requires cudnn5 and the prediction requires the pre-trained model parameters, which is not a small file and may not be suitable to put it into github.
I suggest we remove rnn and vgg. There are examples about rnn and vgg in examples/ folder (no notebooks).
I think, we can add one notebook to show the usage of tensor, initializer, optimizer, devices, and other components before regression/rbm/mlp.

moazreyad: RNN and VGG are removed.

moazreyad:
I checked the notebook and they work fine. I made small fixes and merged all changes in one commit.

nudles:
I moaz, please check again.
You can add new commits (no need to overwrite the existing commits).

moazreyad:
I got an error:

 43     print 'Epoch %d, Reconstruction error per image = %f' % (epoch, err / num_train_batch / batch_size)

NameError: name 'num_train_batch' is not defined

in rbm.ipynb

moazreyad:
Another problem: regression.ipynb points to the Tensor example (http://singa.incubator.apache.org/en/docs/tensor.html) which is not working. It gives an error: 

NameError: name 'np' is not defined

I think the tensor example in the documentation needs to be updated.

PR #281, opened 2016-11-20
SINGA-273 Improve license and contributions

moazreyad:


LICENSE
```diff
@@ -186,7 +186,7 @@ Apache License
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright {yyyy} {name of copyright owner}
```
xiezl: This change is invalid as it should be "[""]" according to https://www.apache.org/licenses/LICENSE-2.0

LICENSE
```diff
@@ -199,287 +199,3 @@ Apache License
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
```
xiezl: If you are going to remove the thirdparty license into independent files, you should add related pointer here.

thirdparty/LICENSE
```diff
@@ -0,0 +1,283 @@
+The SINGA project contains subcomponents with separate copyright
```
xiezl: Since there are multiple licenses in this file, you should separate them into different files. Each of those files includes an individual license.

LICENSE
moazreyad: Please note that in the link that you shared above (http://www.apache.org/dev/licensing-howto.html#bundle-asf-product), it says:

> Dependencies which are not included in the distribution MUST NOT be added to LICENSE and NOTICE. As far as LICENSE and NOTICE are concerned, only bundled bits matter

This means SINGA does not need any third party license because no third party "bits" are actually bundled with it.

thirdparty/LICENSE
moazreyad: I don't think it is required to separate the license in multiple files. See MongoDB third party license for example:

https://github.com/mongodb/mongo/blob/master/distsrc/THIRD-PARTY-NOTICES

moazreyad:
Please note also that [Doxyfile](https://github.com/apache/incubator-singa/blob/master/doc/Doxyfile) can not have comments and must have Apache license header. See http://www.apache.org/legal/resolved.html

> Can we use Doxygen-generated config files?

> As long as the generated comments are removed from the Doxygen-generated files, these files may be used. 

And see for example: https://github.com/apache/avro/blob/master/lang/c%2B%2B/Doxyfile

moazreyad:
The Tinydir license can be safely removed from SINGA, because it is already included in [tinydir.h](https://github.com/apache/incubator-singa/blob/master/include/singa/utils/tinydir.h) .

See https://github.com/cxong/tinydir :
> if you use tinydir you can comply by including tinydir.h

LICENSE
nudles: Hi Moaz,

Please refer to this example created by Apache PMC, https://github.com/justinmclean/ApacheWombat/.
We did bundle some code from other projects, hence we need to include their licenses. We can do in two ways,
1. Add the dependent licenses in the LICENSE file. This is what we are following now.
2. Add the license file of each dependent lib in a separate file and add a link/pointer to that file in LICENSE. Just follow the above example.

nudles:
It would be safe to mention the license of tinydir.h in the the LICENSE file together with others.

moazreyad:
Can you please check now if the PR is ok? 

I still think the license can be further improved, but at least we improved some issues in this PR.

PR #286, opened 2016-12-03
SINGA-278 Convert trained caffe parameters to singa

XiangruiCAI:
Convert trained parameters of caffe model to singa.
Run vgg as an example.


nudles:
To test his PR, please build singa with python support and install it into your virtual env;
 execute  './run.sh'

PR #291, opened 2016-12-15
[WIP] Add xml report for python unittest

nudles:
To run python unittest by Jenkins and display the test results in a xml report.
1. update the docker file to install python package for xml reporting
2. update jenkins script to run python unittest
3. update jenkins setting to collect xml reports.


nudles:
refer to #294

PR #297, opened 2016-12-28

moazreyad:
Script for building binary Debian packages for SINGA 1.0.0 with C++ and with Python for amd64 architecture.


nudles:
@moazreyad Thanks.
1. It would be better to avoid duplicating the source code in the git repo. Otherwise, we have to update the copied version every time we update the original source code. The build.sh file could be updated to do this copy. Then we can call Jenkins to execute the build.sh to generate new packages when new commits are merged.
2. Move the debian folder into tool, just like mesos, docker and jenkins?

moazreyad:
Script updated to build and copy files. It is also moved to 'tool' folder.

tool/debian/singa-cpp_1.0.0_amd64/DEBIAN/control
```diff
@@ -0,0 +1,12 @@
+Package: Singa
+Version: 1.0.0
+Architecture: amd64
+Maintainer: Moaz Reyad <moazreyad@gmail.com>
+Depends: libgoogle-glog-dev, libprotobuf-dev, libstdc++6 , libc6
```
nudles: how about openblas?
I think we need both libprotobuf and libopenblas.
Would apt-get install the 'Depends' automatically if they are not available in the system?

moazreyad: Yes. apt-get should install the needed libraries automatically. This is tested as in [SINGA-274](https://issues.apache.org/jira/browse/SINGA-274). 

p.s. the Debian files assume the build will run on amd64. If the host system used for building the Debian packages is something else, then the Architecture should be changed.

PR #308, opened 2017-02-15

moazreyad:
Build binary Debian package for SINGA GPU (CUDA 8 - cuDNN 5)


tool/debian/singa-cuda8.0-cudnn5/DEBIAN/control
```diff
@@ -0,0 +1,14 @@
+Package: singa-cuda8
+Version: 1.0.1
```
nudles: should be 1.1.0?

tool/debian/build.sh
```diff
@@ -86,26 +86,22 @@ if [ -n "$BUILD_ID" ]; then
 fi
 
 FOLDER=$FOLDER_PREFIX-cpp
-if [ $CUDA = "ON" ]; then
-  FOLDER=$FOLDER_PREFIX-cuda$CUDA_VERSION-cudnn$CUDNN_VERSION
-fi
-
 if [ $PYTHON = "ON" ]
 then
   FOLDER=$FOLDER/python-singa
```
nudles: need to update the path?
the path for python part has been changed from python-singa to singa-cpp/python-singa/

moazreyad: the path is already fixed. no need to update it.

tool/debian/singa-cuda8.0-cudnn5/DEBIAN/control
moazreyad: This number is updated with your build.sh during the building process with the correct version.

nudles: I see. Then please update the top cmake file to change the version ID to 1.1.0.
thanks.

moazreyad: The version ID is set correctly in CMakeLists.txt: SET(PACKAGE_VERSION "1.1.0"). No need to update any thing. The Debian package will be built with the correct version from CMakeLists.txt.

tool/debian/build.sh
nudles: 1. singa-cpp folder has two subfolders： python-singa and singa
2. singa-cuda8.0-cudnn5 does not have the same subfolder.
should we make them consistent by following either 1 or 2?

moazreyad: These directories are only a temporary solution for saving the control files. We do not need them to be perfect. Later if we use CPack, these directories could be removed as the control files will be generated automatically.

nudles: the folder name decides the deb file name.
we'd better make them consistent. otherwise it would be confused if we later add a python-singa-cuda8.0-cudnn5 (the current singa-cuda8.0-cudnn5 does include the python part).
one solution is not differentiating python-singa and singa; 
another solution is differentiating them clearly (I prefer this).

nudles:
I generated the singa-cpp/python-singa.deb using ubuntu:14.04 docker image.
but it cannot be installed on aws with ubuntu14.04. 
the error log is attached.
[out.txt](https://github.com/apache/incubator-singa/files/782297/out.txt)

moazreyad:
To install python-singa.deb on Ubuntu 14:04 :
```
 apt-get update && apt-get install ca-certificates
 dpkg -i ./python-singa.deb 
 apt-get install -f
```
To build  python-singa.deb on Ubuntu 14:04 :

```
apt-get update && apt-get install git cmake  python2.7-dev python-pip python-numpy libprotobuf-dev libopenblas-dev protobuf-compiler libpcre3 libpcre3-dev wget
wget https://launchpad.net/ubuntu/+archive/primary/+files/swig_3.0.2.orig.tar.gz
tar -xvzf ./swig_3.0.2.orig.tar.gz 
cd swig-3.0.2/
./configure 
make
make install
cd ..
git clone https://github.com/moazreyad/incubator-singa.git
cd incubator-singa/
git checkout SINGA-274
./tool/debian/build.sh -b -p
```

moazreyad:
This PR can be closed because [PR 309](https://github.com/apache/incubator-singa/pull/309) provides better solution with CPack.

PR #309, opened 2017-02-20

moazreyad:
Use CPack to generate the Debian packages from the make command:

`make package`


CMakeLists.txt
```diff
@@ -163,3 +163,55 @@ INSTALL(FILES ${CMAKE_BINARY_DIR}/include/singa/singa_config.h DESTINATION
 INSTALL (DIRECTORY ${CMAKE_BINARY_DIR}/lib DESTINATION
     ${CMAKE_INSTALL_PREFIX}
     PATTERN "*libgtest.a" EXCLUDE )
+
+IF (USE_PYTHON)
+INSTALL(DIRECTORY ${PROJECT_SOURCE_DIR}/python
```
nudles: It think python setup.py install can copy these files. Then we don't need to copy them explicitly.
https://github.com/nudles/incubator-singa/blob/conda/CMakeLists.txt#L166

tool/debian/build.sh
```diff
@@ -80,32 +80,25 @@ if [ $BUILD = "ON" ]; then
...
 then
-  FOLDER=$FOLDER/python-singa
+  FOLDER=$FOLDER_PREFIX-cuda$CUDA_VERSION-cudnn$CUDNN_VERSION
 else
-  FOLDER=$FOLDER/singa
+  FOLDER=singa-cpp/$FOLDER_PREFIX
 fi
 
 echo "Path: " build/debian/$FOLDER
 mkdir -p build/debian/$FOLDER
```
nudles: Do we still need these scripts if cpack is supported?

moazreyad: No, we don't need them. I will delete them.

nudles: yes. I think we just need the first few lines and then call cmake and make pakcage.
please also  try to reformat the commit message (title).

CMakeLists.txt
moazreyad: I think we still need `INSTALL(DIRECTORY ${PROJECT_SOURCE_DIR}/python ` for packaging. Otherwise, the directory /python will not be packaged.

tool/debian/build.sh
nudles: sure. we need to create one folder say xxxx and python setup.py install --prefix=xxx. then all python related files would be copied into xxx.

moazreyad: Files deleted and commit message fixed.

nudles:
I got another error:
```
[100%] Built target _singa_wrap
Run CPack packaging tool...
CPack: Create package using DEB
CPack: Install projects
CPack: - Run preinstall target for: singa
CPack: - Install project: singa
CPack: Create package
CPack: - package: /incubator-singa/build/python-singa-cuda-1.1.0.deb generated.
root@001c781d59d5:/incubator-singa/build# ls    
CMakeCache.txt  CPackConfig.cmake        Makefile         bin                  cnmem         examples  install_manifest.txt  python                       src
CMakeFiles      CPackSourceConfig.cmake  _CPack_Packages  cmake_install.cmake  cnmem-prefix  include   lib                   python-singa-cuda-1.1.0.deb  test
root@001c781d59d5:/incubator-singa/build# mv python-singa-cuda-1.1.0.deb /
root@001c781d59d5:/incubator-singa/build# cd /
root@001c781d59d5:/# apt-get install ./python-singa-cuda-1.1.0.deb 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'python-singa-cuda' instead of './python-singa-cuda-1.1.0.deb'
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 python-singa-cuda : Depends: libgoogle-glog-dev but it is not installable
                     Depends: python-numpy but it is not installable
                     Depends: python-pillow but it is not installable
E: Unable to correct problems, you have held broken packages.
```

moazreyad:
Avoid using apt-get and use dpkg instead:

```
apt-get update && apt-get install ca-certificates
dpkg -i ./python-singa-cuda-1.1.0.deb 
apt-get install -f
```

nudles:
```
root@001c781d59d5:/# apt-get update && apt-get install ca-certificates
Get:1 http://archive.ubuntu.com/ubuntu xenial InRelease [247 kB]
Ign:2 http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64  InRelease
Get:3 http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64  Release [564 B]
Get:4 http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64  Release.gpg [819 B]
Get:5 http://archive.ubuntu.com/ubuntu xenial-updates InRelease [102 kB]           
Get:6 http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1604/x86_64  Packages [44.7 kB]
Get:7 http://archive.ubuntu.com/ubuntu xenial-security InRelease [102 kB]
Get:8 http://archive.ubuntu.com/ubuntu xenial/main Sources [1103 kB]
Get:9 http://archive.ubuntu.com/ubuntu xenial/restricted Sources [5179 B]                   
Get:10 http://archive.ubuntu.com/ubuntu xenial/universe Sources [9802 kB]
Get:11 http://archive.ubuntu.com/ubuntu xenial/main amd64 Packages [1558 kB]
Get:12 http://archive.ubuntu.com/ubuntu xenial/restricted amd64 Packages [14.1 kB]
Get:13 http://archive.ubuntu.com/ubuntu xenial/universe amd64 Packages [9827 kB]
Get:14 http://archive.ubuntu.com/ubuntu xenial-updates/main Sources [293 kB]
Get:15 http://archive.ubuntu.com/ubuntu xenial-updates/restricted Sources [2815 B]
Get:16 http://archive.ubuntu.com/ubuntu xenial-updates/universe Sources [164 kB]
Get:17 http://archive.ubuntu.com/ubuntu xenial-updates/main amd64 Packages [615 kB]
Get:18 http://archive.ubuntu.com/ubuntu xenial-updates/restricted amd64 Packages [12.4 kB]
Get:19 http://archive.ubuntu.com/ubuntu xenial-updates/universe amd64 Packages [515 kB]
Get:20 http://archive.ubuntu.com/ubuntu xenial-security/main Sources [72.9 kB]
Get:21 http://archive.ubuntu.com/ubuntu xenial-security/restricted Sources [2392 B]
Get:22 http://archive.ubuntu.com/ubuntu xenial-security/universe Sources [23.0 kB]
Get:23 http://archive.ubuntu.com/ubuntu xenial-security/main amd64 Packages [272 kB]
Get:24 http://archive.ubuntu.com/ubuntu xenial-security/restricted amd64 Packages [12.0 kB]                                                                                                                 
Get:25 http://archive.ubuntu.com/ubuntu xenial-security/universe amd64 Packages [95.5 kB]                                                                                                                   
Fetched 24.9 MB in 6s (3771 kB/s)                                                                                                                                                                           
Reading package lists... Done
Reading package lists... Done
Building dependency tree       
Reading state information... Done
ca-certificates is already the newest version (20160104ubuntu1).
ca-certificates set to manually installed.
0 upgraded, 0 newly installed, 0 to remove and 59 not upgraded.
root@001c781d59d5:/# dpkg -i ./python-singa-cuda-1.1.0.deb 
Selecting previously unselected package python-singa-cuda.
(Reading database ... 19219 files and directories currently installed.)
Preparing to unpack ./python-singa-cuda-1.1.0.deb ...
Unpacking python-singa-cuda (1.1.0) ...
dpkg: dependency problems prevent configuration of python-singa-cuda:
 python-singa-cuda depends on libgoogle-glog-dev; however:
  Package libgoogle-glog-dev is not installed.
 python-singa-cuda depends on python-numpy; however:
  Package python-numpy is not installed.
 python-singa-cuda depends on python-pillow; however:
  Package python-pillow is not installed.

dpkg: error processing package python-singa-cuda (--install):
 dependency problems - leaving unconfigured
Errors were encountered while processing:
```

moazreyad:
This is normal, just ignore the errors and do 

`apt-get install -f`

CMakeLists.txt
nudles: can we package the files in $CMAKE_INSTALL_PREFIX/lib/python/site-package/singa/, which is the place of python files when we type `make install` and is the default folder of python modules?

moazreyad: If you test it and it works, then fine.

moazreyad:
I see that [line 193 in CMakeLists.txt](https://github.com/apache/incubator-singa/blob/master/CMakeLists.txt#L193) is commented:

`SET(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA "${PROJECT_SOURCE_DIR}/tool/debian/postinst;${PROJECT_SOURCE_DIR}/tool/debian/postrm" )`

And also the folder `tool/debian` is deleted. 

Why did you remove this line and folder? CPack for python-singa will work without them.

nudles:
when we execute `make install`, cmake would execute

    python setup.py install

to install python files.
Hence, we don't need the postins and postrm files.
I passed the test for the CPU version. but it failed for the gpu version.

moazreyad:
The packaging scenario is different from the` make install` scenario. In ` make install`, the code is built and installed on the same machine. In the packaging scenario, the code is built on one machine and installed on another. It is true that postins and postrm are not needed if the code is built and installed on the same machine, but they are needed for the Debian packaging. 

`Make` will not be called when installing the binary Debian package with apt (dpkg) and may be `make` will not be available at all on the target machine.

The test will pass for make install, but all python packaging with CPack will not work. The Debian package will not work both for CPU and GPU because at least the `postins` file is needed.

nudles:
when you execute `make package`, would cmake execute `make` and `make install`?
previously, you copied the python files into the prefix folder https://github.com/moazreyad/incubator-singa/blob/cde701abd3ae295505600820708d79d2e471438a/CMakeLists.txt#L171
which is equivalent to `make install`. But we need to add new lines to if the python folder has changes later.
In addition, `make install` would execute `python setup.py install`, which is equivalent to `pip install .` in postinst. 
hence, there is no need for postinst.

moazreyad:
The process of CPack is make > make install > execute install scripts > pack all files to bundle (from [here](http://stackoverflow.com/questions/18722972/prevent-install-code-script-during-cpack-package-generation)). This is executed on the building machine, not on the target machine. 

On the target machine, the package manager (apt) have to do equivalent operations to the `make install `using shell scripts (`postinst`) and copy commands (`INSTALL FILES `and` INSTALL DIRECTORY`). This is why it is required also to copy the python files. If we do not copy them, they will not be found in the target machine if it is different from the build machine which is the expected scenario for the package.

Note: It is better to use `pip install .` than` python setup.py `in the `postinst`.

nudles:
I understand that the building machine and the targeting machines are different.
If the python-singa.deb contains all files (including libsinga.so, the headers and python scripts copied by make install, excludes those from apt-get and pip install), then it should be able to put them into the correct folder of the target machine. 
For the libs from apt-get, it also can install them as they are configured in cpack. 
But for the external python packages, it may not be able to install them automatically. Hence we need a postinst file to just install those files?

moazreyad:
`postinst` is needed to install the SINGA python module (and its dependent external python packages) on the target machine. This is required to run SINGA from python. Otherwise, the user must install it manually.

`postrm` is needed for cleaning if the package is uninstalled. This is required if the user wants to remove the package and install another version. Otherwise, the user must deleted the files manually.

Note: python-singa.deb will not contain all the files if the (INSTALL FILES and INSTALL DIRECTORY) are removed.

nudles:
I think https://github.com/apache/incubator-singa/blob/master/CMakeLists.txt#L172 is equivalent to https://github.com/moazreyad/incubator-singa/blob/cde701abd3ae295505600820708d79d2e471438a/CMakeLists.txt#L171. Both will copy the python scripts into the prefix folder.

moazreyad:
The first one will copy the scripts into the prefix folder in **the build machine**. The second one will copy the scripts into the folder in **the target machine**.

The first one has no effect on the `make package` command. It will be executed, but without any useful effect. It can be removed safely and` make package` will still work fine. (But the `make install` will fail). This is why both are needed.

nudles:
how can the second one be executed on the target machine?
It is just a cmake command, which copies the files into the prefix folder in the build machine.
the first one also copies the files into the prefix folder. hence, I think they are equivalent.

I think cpack just compress all files in a folder into .deb. apt-get would decompress the files and put them into the system folder, and install some dependent libs, and execute postinst.

moazreyad:
I didn't say the second one will be **executed** on the target machine. The second one **will copy file files in the target machine**. This is because INSTALL is not only a cmake command, it is also a cpack command. If the file is not in the INSTALL command, cpack will not copy it. 

In summary: 

the first one is redundant to` make package` but necessary for `make install`
the second one is  redundant to `make install `but necessary for `make package`

So please keep both of them if we want both `make install` and` make package `to work.

nudles:
sorry. I am still confused about "This is because INSTALL is not only a cmake command, it is also a cpack command. If the file is not in the INSTALL command, cpack will not **copy** it."

INSTALL only works on the build machine. I understand that without INSTALL, cpack does not copy those files into the prefix folder. But it doesn't matter, as the first command would copy those files into the prefix folder.

I understand that cpack a tool which is used during generating .deb file.  apt-get is another tool which decompress .deb file and put the files into the system folder. cpack would not be involved when we do `apt-get install ./python-singa.deb`.

As long as all files are in the .deb file, apt-get should put them into the correct folders.

moazreyad:
Yes, I mean it will copy the files to the target machine. CPack will archive the files in the Debian package and the dpkg command on the target machine will unarchive them.

The Debian package must contain setup.py, _singa_wrap.so, singa_wrap.py and the two directories DIRECTORY ${PROJECT_SOURCE_DIR}/python , ${CMAKE_BINARY_DIR}/python/singa/proto to be able to install the SINGA python module. 

If you removed the lines:

```
INSTALL(DIRECTORY ${PROJECT_SOURCE_DIR}/python 
	DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa)	

INSTALL(FILES ${CMAKE_BINARY_DIR}/python/setup.py 
	DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python)
INSTALL(FILES ${CMAKE_BINARY_DIR}/python/singa/singa_wrap.py
	DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa)
INSTALL(FILES ${CMAKE_BINARY_DIR}/python/singa/_singa_wrap.so
	DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa)
INSTALL(DIRECTORY ${CMAKE_BINARY_DIR}/python/singa/proto
	DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa)
```
Then these files will not be in the debian package generated by make package:
```
ar x python-singa-1.1.0.deb
tar -xvzf data.tar.gz | grep setup.py
```
will show nothing, because setup.py is not in the package.

moazreyad:
If you want to eliminate the redundancy of the both cases, then we can use "IF" condition to run only the needed commands in each case:

```
IF (CPACK_GENERATOR MATCHES "DEB")
	IF (USE_PYTHON)
	INSTALL(FILES ${CMAKE_BINARY_DIR}/python/setup.py 
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python)
	INSTALL(FILES ${CMAKE_BINARY_DIR}/python/singa/singa_wrap.py
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa)
	INSTALL(FILES ${CMAKE_BINARY_DIR}/python/singa/_singa_wrap.so
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa)
	INSTALL(DIRECTORY ${PROJECT_SOURCE_DIR}/python 
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa)	
	INSTALL(DIRECTORY ${CMAKE_BINARY_DIR}/python/singa/proto
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa)	
	ENDIF()
ELSE()
	INSTALL(CODE "execute_process(COMMAND python setup.py install --prefix=${CMAKE_INSTALL_PREFIX} WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/python)")
ENDIF()
```

nudles:
OK. I will try it.
Thanks.

moazreyad:
If you want to install the files only with CPack to avoid redundancy during make install, then please try to use these lines:

```
IF (USE_PYTHON)
	INSTALL(FILES ${CMAKE_BINARY_DIR}/python/setup.py 
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python
		COMPONENT PySinga)
	INSTALL(FILES ${CMAKE_BINARY_DIR}/python/singa/singa_wrap.py
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa
		COMPONENT PySinga)
	INSTALL(FILES ${CMAKE_BINARY_DIR}/python/singa/_singa_wrap.so
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa
		COMPONENT PySinga)
	INSTALL(DIRECTORY ${PROJECT_SOURCE_DIR}/python 
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa
		COMPONENT PySinga)	
	INSTALL(DIRECTORY ${CMAKE_BINARY_DIR}/python/singa/proto
		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa
		COMPONENT PySinga)	
	SET(CPACK_COMPONENTS_ALL PySinga)
ENDIF()
```

PR #310, opened 2017-02-20
SINGA-301 Hardcode OpenCL source code as static strings

undisputed-seraphim:
[Reference](https://issues.apache.org/jira/browse/SINGA-301).

- Fixed a bug in OpenCL MemberSign unit test where the expected output should be -1.0, not 0.0, causing the test to fail.
- Python script that converts a cl source file to a std string within a .cpp file.
  The declaration of the strings is fixed as an `extern` in devices.h.


src/core/device/opencl_device.cc
```diff
@@ -93,29 +90,11 @@ void OpenclDevice::CopyDataToFrom(Block* dst, Block* src, size_t nBytes,
...
-	buffer << clFile.rdbuf();
-	std::string clSrc(buffer.str());
-
-	std::string name(file.name);
-    ocl::current_context().add_program(clSrc, name);
-
-	tinydir_next(&dir);
-  }
+void OpenclDevice::BuildPrograms() {
+  ocl::current_context().add_program(distribution_str, "distribution.cl");
```
nudles: at runtime, both .cl and _str.cc files are used?
then, why do we need _str.cc?

PR #312, opened 2017-02-22
SINGA-304 Add precision and recall for multilabel metrics

RUAN0007:
Fix some bug for ffnet

### DIFF
--- a/python/singa/metric.py
+++ b/python/singa/metric.py
@@ -87,65 +87,6 @@ class Accuracy(Metric):
         self.swig_metric = singa.Accuracy()
 
 
-class Precision(Metric):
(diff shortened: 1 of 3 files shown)


python/singa/metric.py
```diff
@@ -78,9 +79,197 @@ def evaluate(self, x, y):
...
-    '''Compute the top one accuracy for singel label prediction tasks.
+    '''Compute the top one accuracy for single label prediction tasks.
 
     It calls the C++ functions to do the calculation.
     '''
     def __init__(self):
         self.swig_metric = singa.Accuracy()
+
+
+class Precision(Metric):
```
kaiping: Why there are two Precision classes here?

PR #317, opened 2017-03-17
SINGA-274 Improve Debian packaging with CPack

moazreyad:
This PR fixes improves #309.


CMakeLists.txt
```diff
@@ -171,6 +171,25 @@ INSTALL(DIRECTORY ${CMAKE_BINARY_DIR}/lib DESTINATION ${CMAKE_INSTALL_PREFIX} PA
...
+	INSTALL(FILES ${CMAKE_BINARY_DIR}/python/setup.py 
+		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python
+		COMPONENT PySinga)
+	INSTALL(FILES ${CMAKE_BINARY_DIR}/python/singa/singa_wrap.py
+		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa
+		COMPONENT PySinga)
+	INSTALL(FILES ${CMAKE_BINARY_DIR}/python/singa/_singa_wrap.so
+		DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/singa/python/singa
+		COMPONENT PySinga)
+	INSTALL(DIRECTORY ${PROJECT_SOURCE_DIR}/python
```
nudles: If this command copies everything in python/ folder, then we don't need to copy the files (setup.py and singa_wrap.py...) one by one.

moazreyad: This command copies the files in the **build machine** not in the **target machine** as we discussed in #309. 

if you build the package with your modification, it will not work because it will be missing the Python files:

```
ar x python-singa-1.1.0.deb
tar -xvzf data.tar.gz 
```

I created this PR to add these lines again as they are deleted in the previous PR. Please do not remove them. If they cause a problem, please let me know what is the problem. The problem can not be that they are redundant because they are required.

nudles: I am not going to remove those lines.
Just want to check if line 184 can copy all files under python/ folder, then we don't need line 175, 178, 181 and 187?

moazreyad: Line 184 copies the Python **source** files `${PROJECT_SOURCE_DIR}/python `. The other four lines copy the Python **binary** files `${CMAKE_BINARY_DIR}/python`. So both of them are needed because this is a binary package not a source only package. 

If we copy only the sources, then the user must install the build tools and compile the Python module by himself.

nudles:
OK. Noted.
Another question, when we do 
```
apt-get install . -f
```
where would the python-singa be installed? under /usr/lib/singa/python/singa? or /usr/lib/python2.7/site-pakcage/singa?

moazreyad:
The python-singa will be installed in the python packages directory. The files in /usr/lib/singa/python/singa are deleted now after the installation. No need for postrm file.

moazreyad:
While working on the build, I noticed that there are two files named rnn.cc:

https://github.com/apache/incubator-singa/blob/master/src/model/layer/rnn.cc
https://github.com/apache/incubator-singa/blob/master/src/model/rnn.cc

The second one is empty and could be removed. 

Can I also remove it as part of this PR?

nudles:
yes.

nudles:
If I do `make & make install`, then the '/usr/local/lib/singa/' folder would exist.
If I create the conda package `conda-build tool/conda`, the 'lib/singa' folder would also be in the conda pacakge.

Can we add a check in the CMakefileLists.txt to copy the files into ${CMAKE_INSTALL_PREFIX}/lib/singa only when we are executing `make package`. As that folder is only used if we install singa via apt-get install ..

moazreyad:
Ok. Check added, but now to build the package the user must set the PACKAGE variable when calling CMake:

`cmake -DPACKAGE=YES ..`

nudles:
like this?
```
cmake -DPACKAGE=YES ..
make package
```

moazreyad:
Yes.

PR #322, opened 2017-05-19
random_size_crop in image_tool.py

Sudhar287:
Crop the image at random offset and get a patch of random size

### DIFF
--- a/python/singa/image_tool.py
+++ b/python/singa/image_tool.py
@@ -32,6 +32,7 @@ Example usage::
 import random
 import numpy as np
 from PIL import Image, ImageEnhance
+import math
 
 
 def load_img(path, grayscale=False):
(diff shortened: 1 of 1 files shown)


python/singa/image_tool.py
```diff
@@ -426,7 +426,33 @@ def random_crop(self, patch, inplace=True):
             return self
         else:
             return new_imgs
-
+
```
nudles: what's difference between random_size_crop and random_crop?

Sudhar287: In random_crop the patch size has to be specified (as an input). This is the output image size.

But in random_size_crop the output size of cropped image is a random integer itself.

nudles: the random_size_crop should be similar to https://github.com/pytorch/vision/blob/master/torchvision/transforms.py#L241

PR #323, opened 2017-05-22
SINGA-290 Upgrade to Python 3

moazreyad:
This upgrade works on Ubuntu > 16.10 without any problem. Both with and without CUDA/cuDNN GPU support.

# For Ubuntu versions before 16.10:

## For building:

You will need protobuf-compiler and libprotobuf-dev > 3.0. So either:

1. Use -DUSE_MODULES=YES while running cmake to build protobuf 3 from source. To build protobuf 3, you will need:

```
apt-get install curl unzip dh-autoreconf
cmake -DUSE_MODULES=YES ..
```

2. Or add new Ubuntu repository to /etc/apt/sources.list before building. For example: 

```
echo "deb http://cz.archive.ubuntu.com/ubuntu yakkety main" >> /etc/apt/sources.list
apt-get update
cmake ..
```

## For installing from Debian package:

If you used the second method to build a package with `make package `, and you want to install the package on Ubuntu < 16.10, then the system must have libprotobuf-dev > 3.0. Use this commands to install the package on Ubuntu < 16.10:

```
echo "deb http://cz.archive.ubuntu.com/ubuntu yakkety main" >> /etc/apt/sources.list
apt-get update
apt-get install  libprotobuf-dev=3.0.0-7ubuntu3
apt-get install ./python3-singa.deb (or python3-singa-cuda.deb)
```

you may replace http://cz.archive.ubuntu.com/ubuntu with another mirror

# Note:

This upgrade is not compatible with Python 2 because the dependencies are different. If we want to support both Python 2 and Python3, then we must add another parameter for cmake, for example:

`cmake -DPYTHON2=YES`


nudles:
If we merge this PR, then we cannot use PY2 anymore..
We can temporarily put it into a new branch. 
But before that, we need a plan on how to merge it with the master later.
For example, can we add some conditional checks in the python code like

    if py2:
       import ConfigParser
    else:
       import configparser

moazreyad:
I used the `future` library and add the `PYTHON2 ` option to `cmake `.

Now both Python 2 and 3 are supported.

python/singa/data.py
```diff
@@ -97,7 +101,7 @@ def start(self):
         self.p.start()
         return
 
-    def next(self):
+    def __next__(self):
```
nudles: next() is called explicitly.
if changed, there would be errors.

python/singa/image_tool.py
```diff
@@ -384,7 +388,7 @@ def crop8(self, patch, num_case=1, inplace=True):
         if num_case == patch5 + patch3:
             count = patch5
         else:
-            sample_list = range(0, patch5 + patch3)
+            sample_list = list(range(0, patch5 + patch3))
```
nudles: according to http://python-future.org/compatible_idioms.html
we can avoid adding a list for range by

    from builtins import range
    # or
    from past.builtins import range

python/singa/net.py
```diff
@@ -321,7 +328,7 @@ def forward(self, flag, x, output=[]):
         # print output_of_layer
         ret.update(output_of_layer)
         if len(ret) == 1:
-            return ret.values()[0]
+            return list(ret.values())[0]
```
nudles: why adding list()?

python/singa/command.py
```diff
@@ -21,12 +21,16 @@
...
 
+from future import standard_library
+standard_library.install_aliases()
+from builtins import str
 import sys, glob, os, random, shutil, time
 from flask import Flask, request, redirect, url_for
 import numpy as np
-import ConfigParser
-import urllib, traceback
+import configparser
```
moazreyad: This code is generated by python-future. See:

http://python-future.org/compatible_idioms.html#configparser

Does it cause a problem or something?

python/singa/net.py
```diff
@@ -53,13 +53,20 @@
...
-from __init__ import __version__
-import tensor
-import layer
-import snapshot
-import cPickle as pickle
+from .__init__ import __version__
+from . import tensor
+from . import layer
+from . import snapshot
+import pickle as pickle
```
moazreyad: In Python 3, cPickle is renamed to pickle.

This code is generated by python-future which has a limitation : 

"We don't currently support these modules
 import pickle     # should (optionally) bring in cPickle on Python 2"

http://python-future.org/_modules/future/standard_library.html

python/singa/image_tool.py
moazreyad: `from past.builtins import range` can not be used because it is for` backward compatible` when converting from Python 3 to Python 2. 

`from builtins import range` is already used in this file. It adds a list for range:

http://python-future.org/compatible_idioms.html#range

python/singa/data.py
moazreyad: python-future handles this:

http://python-future.org/what_else.html#custom-iterators

There should be no errors. If you find an error, please report it.

python/singa/net.py
moazreyad: It is added by python-future for upgrading to Python 3.

moazreyad:
This upgrade supports both Python 2 and Python 3, with and without CUDA/cuDNN GPU support.

The default is Python 3. To build the Python 2 version, set the PYTHON2 option in cmake:

```
cmake -DPYTHON2=YES 
```

To use Python 3, it is required to have SWIG 3.0.10. It is available in Ubuntu >17.04. For other operating systems, you may install it manually or from the Ubuntu 17.04 repository:
```
echo "deb http://archive.ubuntu.com/ubuntu zesty universe" >> /etc/apt/sources.list
apt-get update && apt-get install swig3.0
```

examples/cifar10/predict.py
```diff
@@ -52,7 +52,7 @@ def predict(net, images, dev, topk=5):
 def load_dataset(filepath):
     print('Loading data file %s' % filepath)
     with open(filepath, 'rb') as fd:
-        cifar10 = pickle.load(fd)
+        cifar10 = pickle.load(fd, encoding='latin1')
```
nudles: do we need to use latin1 if the file was dumped using pickle without specify the encoding field?

moazreyad: Yes, we need latin1 for loading the files in cifar10 example. See :

https://stackoverflow.com/questions/28218466/unpickling-a-python-2-object-with-python-3

python/singa/net.py
nudles: since future does not handle the pickle package properly, i am going to use the idiom 

```
try:
   import pickle
except ImportError:
   import cPickle as pickle
```

python/singa/command.py
nudles: in py2 configparser comes with python.
ok. so now we just add it as an dependent lib.

python/singa/converter.py
```diff
@@ -226,4 +230,4 @@ def convert_params(self, net):
                 i += 1
                 params[i].copy_from_numpy(bias)
                 i += 1
-                print 'converting layer {0}, wmat shape = {1}, bias shape = {2}'.format(layer.name, w.shape, bias.shape)
+                print('converting layer {0}, wmat shape = {1}, bias shape = {2}'.format(layer.name, w.shape, bias.shape))
```
nudles: can py2 support `print('converting layer {0}'.format())`

nudles: seems it is supported since py2.6

examples/mnist/train.py
```diff
@@ -33,16 +39,16 @@
 
 def load_train_data(file_path):
     f = gzip.open(file_path, 'rb')
-    train_set, valid_set, test_set = cPickle.load(f)
+    train_set, valid_set, test_set = pickle.load(f)
```
moazreyad: Yes. it must be `load(f, encoding='latin1')`

nudles:
I have passed the test for conda building.
Shall we use  python 2 as the default option and use -DPYTHON3=ON for python 3?
It seems that package x is for python2 by default, and x-python3 is for python 3.

moazreyad:
The support of Python 2 will end in 2020 (See https://pythonclock.org), so I would say it is better to use Python 3 as the default building option and encourage users to avoid building new solutions with Python 2 unless there is a good reason. 

But this is just a design choice. If most of the current users will build with Python 2, then you can set it as the default for now.

In any case, the naming of the package is independent of which version is the default.

nudles:
Is it necessary to require swig with version>= 3.0.10? which was set to >=3.0.

moazreyad:
Yes, for Python 3, swig 3.0.10 is required.

nudles:
Did this error appear for py2 using swig-3.0.10,

```
ImportError: Failed to import test module: test_metric
  File "/usr/lib/python2.7/unittest/loader.py", line 254, in _find_tests
    module = self._get_module_from_name(name)
  File "/usr/lib/python2.7/unittest/loader.py", line 232, in _get_module_from_name
    __import__(name)
  File "/var/jenkins/workspace/singa-linux-gpu/label/ubuntu14.04-cuda8-cudnn5-conda/lang/CUDA/test/python/test_metric.py", line 23, in <module>
    from singa import metric
  File "/var/jenkins/workspace/singa-linux-gpu/label/ubuntu14.04-cuda8-cudnn5-conda/lang/CUDA/build/python/singa/__init__.py", line 18, in <module>
    from . import singa_wrap
  File "/var/jenkins/workspace/singa-linux-gpu/label/ubuntu14.04-cuda8-cudnn5-conda/lang/CUDA/build/python/singa/singa_wrap.py", line 123
    def value(self) -> "PyObject *":
                    ^
SyntaxError: invalid syntax
```
The compilation and c++ tests work well. but the python tests failed due to this error.

PR #330, opened 2017-06-04
random crop then resize

Sudhar287:
random corp the image then resize. 
similar to https://github.com/pytorch/vision/blob/master/torchvision/transforms.py#L330

### DIFF
--- a/python/singa/image_tool.py
+++ b/python/singa/image_tool.py
@@ -32,6 +32,7 @@ Example usage::
 import random
 import numpy as np
 from PIL import Image, ImageEnhance
+import math
 
 
 def load_img(path, grayscale=False):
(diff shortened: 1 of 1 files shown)


python/singa/image_tool.py
```diff
@@ -115,6 +116,11 @@ def crop_and_resize(img, patch, position):
     # print "crop to box %d,%d,%d,%d and scale to %d,%d" % (box+crop)
     return new_img
 
+def resize_tuple(img,patch):
```
nudles: this function is too simple to exist.
you can call `img.resize(path)` directly.
instead of calling `resize_tuple`.

python/singa/image_tool.py
```diff
@@ -431,6 +437,41 @@ def random_crop(self, patch, inplace=True):
...
+            area=img.size[0]*img.size[1]
+            target_area = random.uniform(0.08, 1.0) * area
+            aspect_ratio = random.uniform(3. / 4, 4. / 3)
+            crop_x = int(round(math.sqrt(target_area * aspect_ratio)))
+            crop_y = int(round(math.sqrt(target_area / aspect_ratio)))
+            assert img.size[0] >= patch[0] and img.size[1] >= patch[1],\
+                'img size (%d, %d), patch size (%d, %d)' % \
+                (img.size[0], img.size[1], patch[0], patch[1])
+            left_offset = random.randint(0, img.size[0] - crop_x)
+            print left_offset
```
nudles: do no print the numbers here.
this function would be called >10000 times

PR #334, opened 2017-06-06
SINGA-323 Add method to resize images by given width and height

XiangruiCAI:
Add methods to resize images by given width and height, named as
resize_by_hw, resize_by_hw_list, resize_by_hw_range;
Encapsulate original "resize" methods into "scale" methods.


python/singa/image_tool.py
```diff
@@ -120,14 +121,26 @@ def resize(img, small_size):
...
     return new_img
 
 
+def scale(img, small_size):
+    '''Scale the image to make the smaller side be at the give size
+        This method is to replace original resize'''
+    return resize(img, small_size)
+
+
+def resize_by_hw(img, (width, height)):
```
nudles: better change the order of width and height to be consistent with the name 'hw'.

python/singa/image_tool.py
```diff
@@ -264,6 +295,45 @@ def resize_by_list(self, size_list, num_case=1, inplace=True):
         else:
             return new_imgs
 
+    def resize_by_hw_range(self, rng, inplace=True):
```
nudles: are resize_by_hw_range and resize_by_hw_list different to scale_by_range and scale_by_list?

XiangruiCAI: Yes. resize_by_hw_range and resize_by_hw_list resize images to _(height, width)_, while scale_by_range and scale_by_list scale _the short edge to a small_size_.

python/singa/image_tool.py
```diff
@@ -121,14 +122,26 @@ def resize(img, small_size):
...
     return new_img
 
 
+def scale(img, small_size):
+    '''Scale the image to make the smaller side be at the give size
+        This method is to replace original resize'''
+    return resize(img, small_size)
+
+
+def resize_by_hw(img, (height, width)):
```
XiangruiCAI: I have changed to _height_ and _width_.

chrishkchris:
it is just some new python method that does not affect the old test case.

All the old test case can pass, so it can merge

PR #338, opened 2017-07-04
SINGA-327 Fix the bug of loading model saved in bin file

dbxinj:
move the version outside if else statement


moazreyad:
By the way, I faced this problem while upgrading to Python 3 and solved it with this [PR](https://github.com/apache/incubator-singa/pull/323/commits/2e93f327c713ab288b94f286d60e2878286acf02):

PR #346, opened 2017-12-14
[SINGA-290] Fix SWIG version for Python 2

moazreyad:
For Python 2, the minimal version of SWIG should be 3.0.8


nudles:
It failed to work with 3.0.8. Hence, I set it to 3.0.10..

PR #350, opened 2018-03-07
SINGA-341 Stride Field Tensor.h test edit

vaanforz:
Modest changes to tensor.h file to check whether am I in right direction

### DIFF
--- a/include/singa/core/tensor.h
+++ b/include/singa/core/tensor.h
@@ -59,7 +59,10 @@ class Tensor {
   explicit Tensor(Shape &&shape, DataType dtype = kFloat32);
   explicit Tensor(const Shape &shape, DataType dtype = kFloat32);
   Tensor(Shape &&shape, std::shared_ptr<Device> dev, DataType dtype = kFloat32);
+  Tensor(Shape &&shape, vector<int> strides, std::shared_ptr<Device> dev, DataType dtype = kFloat32;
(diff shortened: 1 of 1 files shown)


include/singa/core/tensor.h
```diff
@@ -59,8 +59,11 @@ class Tensor {
   explicit Tensor(Shape &&shape, DataType dtype = kFloat32);
   explicit Tensor(const Shape &shape, DataType dtype = kFloat32);
   Tensor(Shape &&shape, std::shared_ptr<Device> dev, DataType dtype = kFloat32);
+  Tensor(Shape &&shape, vector<int> strides, std::shared_ptr<Device> dev, DataType dtype = kFloat32;
```
nudles: by default the strides are computed based on the shape.
e.g. stride[0]=size/shape[0], stride[1]=size/shape[0]/shape[1], ... stride[-1]=1;
hence, we do not need the users to specify strides in the constructor.

include/singa/core/tensor.h
```diff
@@ -210,13 +214,14 @@ class Tensor {
   float L2() const;
 
  protected:
-  bool transpose_ = false;
+  //bool transpose_ = false;
```
nudles: since transpose_ is deleted, you need to update the transpose function.

PR #352, opened 2018-03-20
SINGA-342 Support autograd MLP Example

xuewanqi:


autograd_test1.py
```diff
@@ -0,0 +1,602 @@
...
+        return output_idx#返回prev_fn的为arg_id输出的下标
+
+
+    def _is_ready_for_backward(self, dependencies, function):
+        for deps in dependencies[function]: #list
+            if len(deps) > 0:
+                return False
+        return True
+
+    def run_backward(self, variable, grad):
```
nudles: https://github.com/HIPS/autograd/blob/master/autograd/core.py#L17

The backward (and dependency sort) in autograd is very clear.
what you need is just to count the available output tensors. if all output tensors are ready, then you can do backward.

PR #357, opened 2018-04-19
SINGA-349 Create layer operations for autograd

xuewanqi:
Add more layer operations such as convolution(), pooling() to support autograd.


python/singa/convolution_operation.py
```diff
@@ -0,0 +1,158 @@
+from singa import tensor
```
nudles: rename this file to operation.py? or ops.py

python/singa/convolution_operation.py
```diff
@@ -0,0 +1,158 @@
...
+    // For test use.
+
+
+    To be used in SoftMax Operation.
+    Convert a singa_tensor to numpy_tensor.
+    '''
+    np_array = x.GetFloatValue(int(x.Size()))
+    return np_array.reshape(x.shape())
+
+class Convolution2D(tensor.Operation):
```
nudles: rename it to Conv2D?

python/singa/convolution_operation.py
```diff
@@ -0,0 +1,158 @@
...
+    def __init__(self, name, nb_kernels, kernel=3, stride=1, border_mode='same',
```
nudles: follow the keras API or pytorch API?

python/singa/convolution_operation.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+    def __call__(self, x):
+        if not self.PyLayer.has_setup:
+            self.PyLayer.setup(x.shape[1:])
+        param_data = self.PyLayer.layer.param_values()
+
+        if not hasattr(self, 'w'):
+            self.w = tensor.Tensor(data=param_data[0], requires_grad=True, stores_grad=True)
+            self.w.gaussian(0.0, 0.1)  # TODO realize other initialization method according to W_specs
```
nudles: just use xavier or msra initialization following pytorch.

python/singa/convolution_operation.py
```diff
@@ -0,0 +1,158 @@
...
+
+        xs = [x, self.w]
+
+        if len(param_data) == 2:
+            self.b = tensor.Tensor(data=param_data[1], requires_grad=True, stores_grad=True)
+            self.b.set_value(0.0)  # TODO realize other initialization method according to b_specs
+            xs.append(self.b)
+
+        xs = tuple(xs)
+        return self._do_forward(*xs)
```
nudles: call super class' __call__ method?

python/singa/convolution_operation.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def forward(self, *xs):
+        return self.PyLayer.layer.Forward(4, xs[0])  #how ktrain works?  flag & ktrain.
```
nudles: there seems to be a global var in keras to indicate the phase (training or evaluation)

python/singa/layer_ops.py
```diff
@@ -0,0 +1,171 @@
+from singa import tensor
+from singa import layer
```
nudles: move all implementation related to operations into autograd.py

examples/autograd/mnist_cnn.py
```diff
@@ -0,0 +1,104 @@
...
+    x_test=preprocess(test[0])
+    y_test=to_categorical(test[1],num_classes)
+    print ('the shape of training data is',x_train.shape)
+    print ('the shape of training label is',y_train.shape)
+    print ('the shape of testing data is', x_test.shape)
+    print ('the shape of testing label is', y_test.shape)
+
+    # operations initialization
+    conv1=autograd.Conv2d(3,32)
+    relu1 = autograd.ReLU_Layer()  # same name for tensor.ReLU and layer_ops.ReLU
```
nudles: delete tensor.ReLU?

nudles: you don't need to create the relu1.
just call the operation on the fly.

examples/autograd/mnist_cnn.py
```diff
@@ -0,0 +1,104 @@
...
+    conv2=autograd.Conv2d(32,32)
+    relu2 = autograd.ReLU_Layer()
+    pooling = autograd.MaxPool2d()
+    flatten = autograd.Flatten()
```
nudles: no need to create the flatten/softmax operations, just call it on the fly.

examples/autograd/mnist_cnn.py
```diff
@@ -0,0 +1,104 @@
...
+    linear = autograd.Linear(None, 10)  # in_feature=None for input_shape auto calculation
+    softmax = autograd.SoftMax()
+    cross_entropy = autograd.CrossEntropy()
+
+
+    def forward(x,t):
+        y=conv1(x)[0]
```
nudles: [0] is very ugly. pls try to fix it.

python/singa/autograd.py
```diff
@@ -1,6 +1,531 @@
...
+            dy (CTensor): data for the dL / dy, L is the loss
+
+        Returns:
+            dx (Ctensor): data for the dL / dx, L is the loss,
+            x is the input of current Opertion
+        '''
+        # calculations are made on numpy array
+        if self.axis == 1:
+            dy = dy.T()
+        grad = ctensor2numpy(dy)
```
nudles: the enisum() is supported in singa

python/singa/autograd.py
```diff
@@ -1,6 +1,531 @@
...
+        else:
+            self.flag = model_pb2.kEval
+
+        if not self.PyLayer.has_setup:
+            self.PyLayer.setup(x.shape[1:])
+
+        param_data = self.PyLayer.layer.param_values()
+        if not hasattr(self, 'w'):
+            self.w = Tensor(data=param_data[0], requires_grad=True, stores_grad=True)
+            if self.W_specs['init'] == 'gaussian':
```
nudles: why not call pylayer to do the setup?

python/singa/autograd.py
```diff
@@ -1,6 +1,531 @@
...
+                elif self.b_specs['init'] == 'constant':
+                    self.b.set_value(0.0)
+
+            xs.append(self.b)
+
+        xs = tuple(xs)
+        return self._do_forward(*xs)
+
+    def forward(self, *xs):
+        return self.PyLayer.layer.Forward(self.flag, xs[0])
```
nudles: why not call self.PyLayer.forward()

python/singa/autograd.py
```diff
@@ -1,6 +1,531 @@
...
+        for kwarg in kwargs:
+            if kwarg not in allowed_kwargs:
+                raise TypeError('Keyword argument not understood:', kwarg)
+            else:
+                inner_params[kwarg] = kwargs[kwarg]
+
+        self.W_specs = W_specs
+        self.b_specs = b_specs
+
+        self.PyLayer = layer.Dense(inner_params['name'], num_output=out_features, use_bias=bias,
```
nudles: since you have matmul and add_bias, why not use them?

python/singa/autograd.py
xuewanqi: self.PyLayer.forward return a PyTensor, I need a CTensor to be returned here.

python/singa/autograd.py
xuewanqi: pylarer seems cannot assign variables with real data

python/singa/autograd.py
```diff
@@ -1,6 +1,558 @@
...
+            pad = None
+        else:
+            pad = padding
+
+        if dilation != 1 or groups != 1:
+            raise ValueError('Not implemented yet')
+
+        self.PyLayer = layer.Conv2D(inner_params['name'], nb_kernels=out_channels, kernel=kernel_size, stride=stride,
+                                    border_mode=inner_params['border_mode'],
+                 cudnn_prefer=inner_params['cudnn_prefer'], workspace_byte_limit=inner_params['workspace_byte_limit'],
```
nudles: why not config the args directly

    cudnn_prefer=' fastest' ,
    border_mode='same',

python/singa/autograd.py
```diff
@@ -1,6 +1,558 @@
...
+
+            xs.append(self.b)
+
+        xs = tuple(xs)
+        return self._do_forward_0(*xs)
+
+    def _do_forward_0(self, *xs):
+        return self._do_forward(*xs)[0]
+
+    def forward(self, *xs):
```
nudles: when will forward() be called?

python/singa/autograd.py
nudles: CTensor is just Tensor.data

python/singa/autograd.py
```diff
@@ -1,6 +1,558 @@
...
+        assert type(flag) is bool, 'flag can only be bool.'
+        if flag:
+            self.flag = model_pb2.kTrain
+        else:
+            self.flag = model_pb2.kEval
+
+        if not self.PyLayer.has_setup:
+            self.PyLayer.setup(x.shape[1:])
+
+        return self._do_forward(x)
```
nudles: no need to call _do_forward_0()?

xuewanqi: _do_forward_0() is design to avoid unfriendly API [0]. in padding operation, def max_pool_2d is implemented for this use. so no longer need to call _do_forward_0. 
in fact, other operations usually use def operation to avoid [0]. 
conv operation has its peculiarity, it cannot be easily packaged in that way.

python/singa/autograd.py
xuewanqi: yes. 
the arguments of opeation:forward(), operation:backward() are designed to be Ctensor.
the arguments of layer:forward(), layer:backward() are designed to be Pytensor.
if call layer:forward, Ctensor should be firstly converted to Pytensor, then call layer:forward and return Pytensor, finally convert Pytensor to Ctensor.

python/singa/autograd.py
xuewanqi: forward() is called in function _do_forward().
_do_forward mainly realize two effects:
 1. record and manage source of  a operation 
2. execute calculation by call forward(). 
for _do_foward(), the inputs and outputs should be Pytensor. and the arguments of forward()  are designed to be Ctensor. 
everything about calculation are just did on Ctensor.

python/singa/autograd.py
xuewanqi: yes.
inner_pamas is to provide ports to realize those functions singa_layers have but Pytorch operations don't have.
they can be closed of course.

python/singa/autograd.py
```diff
@@ -1,5 +1,447 @@
...
+            return (self.w, self.b)
+        else:
+            return self.w
+
+    def __call__(self, x, flag=True):
+        assert type(flag) is bool, 'flag can only be bool.'
+        self.flag = flag
+        if self.init_value is False:
+            std = math.sqrt(2.0 / (self.in_features + self.out_features))
+            self.w.gaussian(0.0, std)
```
nudles: put the init steps into the __init__()?

python/singa/autograd.py
```diff
@@ -1,5 +1,447 @@
...
+            if hasattr(self, 'b'):
+                self.b.set_value(0.0)
+            self.init_value = True
+        return self._do_forward(x)
+
+    def _do_forward(self, x):
```
nudles: move the matmul and add_bias into __call__() and delete _do_forward()?

python/singa/autograd.py
xuewanqi: i think its may hard to do so unless realize an alternative setup method for CLayer by sending kernel size.
if put the init into init(), i am not sure if the parameters can connect with CLayer, it must refer the setup codes for Clayer to find how to setup CLayer from outside.

python/singa/autograd.py
xuewanqi: yes, it can.

python/singa/autograd.py
```diff
@@ -1,5 +1,447 @@
...
+        ret = self.PyLayer.layer.Backward(self.flag, dy)
+        return (ret[0],)+ret[1]
+
+
+class Linear(Operation):
+    def __init__(self, in_features, out_features, bias=True):
+        self.in_features = in_features
+        self.out_features = out_features
+        w_shape = (in_features, out_features)
+        self.w = Tensor(shape=w_shape, requires_grad=True, stores_grad=True)
```
nudles: you can init w right here?

python/singa/autograd.py
xuewanqi: can easily put init steps into init() here. i wrongly think that you are indicating conv2d as i just look part of codes. and the last comment is for conv2d.

python/singa/autograd.py
xuewanqi: it can. i design that do default init in __call__, and do particular init by taking out params(call get_params()). 
i can move default init to __init__() here.

PR #363, opened 2018-05-06
SINGA-294 Add instructions for building on Windows

moazreyad:


Wentong-DST:
I will translate the new windows installation guide to Chinese after merging.

PR #365, opened 2018-05-07
SINGA-352 Some fixes in documentation

moazreyad:


nudles:
can we hide the buttons for the past release on the navigation view?

PR #381, opened 2018-05-17
SINGA-369 fix the errors in examples

Sheyujian:
1.fix the error of "Tensor object has no attribute singa_tensor"


python/singa/net.py
```diff
@@ -491,7 +491,7 @@ def get_name(name):
                     f = f + '.pickle'
             assert os.path.exists(f), 'file not exists %s w/o .pickle' % f
             with open(f, 'rb') as fd:
-                params = pickle.load(fd)
+                params = pickle.load(fd,encoding='iso-8859-1')
```
nudles: Which version of python did you use to do the test, Py2 or Py3?

PR #387, opened 2018-05-31
SINGA-371 Implement functional operations in c++ for autograd

xuewanqi:
Functions for convolution operations

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -88,7 +88,7 @@ class Operation(object):
             ys = (ys,)
         # create Tensor based on CTensor(data);
         # assume outputs are all Tensor instances
-        ys = tuple(Tensor(device=y.device,
+        ys = tuple(Tensor(device=y.device(),
                           data=y,
                           requires_grad=self.requires_grad,
                           creator=self) for y in ys)
@@ -369,140 +369,6 @@ def ctensor2numpy(x):
     return np_array.reshape(x.shape())
 
 
-class Conv2d(Operation):
-
-    def __init__(self, in_channels, out_channels, kernel_size=3, stride=1,
-                 padding=0, dilation=1, groups=1, bias=True, **kwargs):
-
-        inner_params = {'name': 'Conv2d',
-                        'border_mode': 'same',
-                        'cudnn_prefer': 'fastest',
-                        'workspace_byte_limit': 1024,
-                        'data_format': 'NCHW',
-                        'W_specs': {'init': 'xavier'},
-                        'b_specs': {'init': 'constant'},
-                        'input_sample_shape': None}
-        # TODO valid value of inner_params check
-
-        for kwarg in kwargs:
-            if kwarg not in inner_params:
-                raise TypeError('Keyword argument not understood:', kwarg)
-            else:
-                inner_params[kwarg] = kwargs[kwarg]
-
-        self.in_channels = in_channels
-        self.out_channels = out_channels
-        self.W_specs = inner_params['W_specs']
-        self.b_specs = inner_params['b_specs']
-
-        if isinstance(kernel_size, int):
-            self.kernel_size = (kernel_size, kernel_size)
-        else:
-            self.kernel_size = kernel_size
-
-        if padding == 0:
-            pad = None
-        else:
-            pad = padding
-
-        if dilation != 1 or groups != 1:
-            raise ValueError('Not implemented yet')
-
-        self.PyLayer = layer.Conv2D(inner_params['name'],
-                                    nb_kernels=out_channels,
-                                    kernel=kernel_size,
-                                    stride=stride,
-                                    border_mode=inner_params['border_mode'],
-                                    cudnn_prefer=inner_params['cudnn_prefer'],
-                                    workspace_byte_limit=inner_params[
-                                        'workspace_byte_limit'],
-                                    data_format=inner_params['data_format'],
-                                    use_bias=bias,
-                                    W_specs=self.W_specs,
-                                    b_specs=self.b_specs,
-                                    pad=pad,
-                                    input_sample_shape=inner_params['input_sample_shape'])
-
-    def get_params(self):
-        assert self.init_value is True, 'must initialize before get_params()'
-        if self.bias:
-            return (self.w, self.b)
-        else:
-            return self.w
-
-    def __call__(self, x):
-        if training:
-            self.flag = model_pb2.kTrain
-        else:
-            self.flag = model_pb2.kEval
-
-        if not self.PyLayer.has_setup:
-            self.PyLayer.setup(x.shape[1:])
-
-        param_data = self.PyLayer.layer.param_values()
-
-        if not hasattr(self, 'w'):
-            self.w = Tensor(device=param_data[0].device, data=param_data[
-                            0], requires_grad=True, stores_grad=True)
-            std = math.sqrt(
-                2.0 / (self.in_channels * self.kernel_size[0] * self.kernel_size[1] + self.out_channels))
-            self.w.gaussian(0.0, std)
-
-        xs = [x, self.w]
-
-        if len(param_data) == 2:
-            if not hasattr(self, 'b'):
-                self.b = Tensor(device=param_data[1].device, data=param_data[
-                                1], requires_grad=True, stores_grad=True)
-                self.b.set_value(0.0)
-
-            xs.append(self.b)
-
-        xs = tuple(xs)
-        return self._do_forward(*xs)[0]
-
-    def forward(self, *xs):
-        return self.PyLayer.layer.Forward(self.flag, xs[0])
-
-    def backward(self, dy):
-        ret = self.PyLayer.layer.Backward(self.flag, dy)
-        return (ret[0],) + ret[1]
-
-
-class Linear(Operation):
-
-    def __init__(self, in_features, out_features, bias=True):
-        self.in_features = in_features
-        self.out_features = out_features
-        self.w_shape = (in_features, out_features)
-        self.b_shape = (1, out_features)
-        self.bias = bias
-        self.init_value = False
-
-    def get_params(self):
-        assert self.init_value is True, 'must initialize before get_params()'
-        if self.bias:
-            return (self.w, self.b)
-        else:
-            return self.w
-
-    def __call__(self, x):
-        if self.init_value is False:
-            self.w = Tensor(shape=self.w_shape,
-                            requires_grad=True, stores_grad=True)
-            std = math.sqrt(2.0 / (self.in_features + self.out_features))
-            self.w.gaussian(0.0, std)
-            if self.bias:
-                self.b = Tensor(shape=self.b_shape,
-                                requires_grad=True, stores_grad=True)
-                self.b.set_value(0.0)
-            self.init_value = True
-        y = matmul(x, self.w)
-        if self.bias:
-            y = add_bias(y, self.b, axis=0)
-        return y
-
-
 class MaxPool2d(Operation):
 
     def __init__(self, kernel_size=3, stride=1, padding=0, dilation=1,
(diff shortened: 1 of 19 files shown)


src/model/convolution functions.cpp
```diff
@@ -0,0 +1,398 @@
...
+            height_,
+            width_,
+            conv_height_,
+            conv_width_,
+            batchsize,
+    };
+
+}
+
+Tensor CudnnConvForward(Tensor x, Tensor W, Tensor b, const ConvHandle ch, const CudnnConvHandle cch){
```
nudles: const Tensor& x, const Tensor& W, ...

src/model/convolution functions.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    size_t height_;
+    size_t width_;
+    size_t conv_height_;
+    size_t conv_width_;
+    size_t batchsize;
+};
+
+// Done in conv2d.__init__()
+ConvHandle SetupConv(const size_t in_channels, const LayerConf &conf){
```
nudles: will this function be called by cpp and cudnn conv layers?

xuewanqi: yes, but there are two  parameters, workspace_byte_limit and prefer_, unnecessary for cpp. this can be refined later. i want firstly let the cudnn part workable.

src/model/convolution_forward.cc
```diff
@@ -0,0 +1,404 @@
...
+    size_t num_filters_;
+
+    bool bias_term_;
+
+    size_t workspace_byte_limit_;
+    std::string prefer_;
+};
+
+
+struct CudnnConvHandle{
```
zmeihui: why not make CudnnConvHandler as subclass of ConvHandler?

src/model/convolution_forward.cc
```diff
@@ -0,0 +1,404 @@
...
+            cudnnAddTensor(ctx->cudnn_handle, &alpha, cch.bias_desc_,
+                           bblock->data(), &beta, cch.y_desc_,
+                           outblock->mutable_data());
+        }, {output.block(), b.block()}, {output.block()});
+    }
+    return output;
+};
+
+// input Tensor W for Reset dW purpose, can avoid this later.
+Tensor CudnnConvBackwardW(const Tensor dy, const Tensor x, const Tensor W, const CudnnConvHandle cch){
```
zmeihui: use reference
`const Tensor&`

python/singa/autograd.py
```diff
@@ -583,6 +583,105 @@ def backward(self, dy):
...
+        w_shape = (self.out_channels, self.in_channels, self.kernel_size[0], self.kernel_size[1])
+        self.W = Tensor(shape=w_shape, requires_grad=True, stores_grad=True)
+        std = math.sqrt(
+                2.0 / (self.in_channels * self.kernel_size[0] * self.kernel_size[1] + self.out_channels))
+        self.W.gaussian(0.0, std)
+
+        if self.bias:
+            b_shape = (self.out_channels,)
+        else:
+            b_shape = (1,) #to keep consistency when to do forward.
```
nudles: no need to create b if bias is False?

python/singa/autograd.py
```diff
@@ -583,6 +583,105 @@ def backward(self, dy):
...
+        self.b = Tensor(shape=b_shape, requires_grad=True, stores_grad=True)
+        self.b.set_value(0.0)
+
+
+    def __call__(self, x):
```
nudles: can move the setup into forward()?

src/model/convolution_functions.cc
```diff
@@ -0,0 +1,297 @@
...
+            height_,
+            width_,
+            conv_height_,
+            conv_width_,
+            batchsize,
+    };
+};
+
+Tensor CudnnConvForward(const Tensor &x, const Tensor &W, const Tensor &b,
+                        const ConvHandle ch, const CudnnConvHandle cch){
```
nudles: why not merge ConvHandle and CudnnConvHandle?

src/model/convolution_functions.h
```diff
@@ -0,0 +1,67 @@
...
+    size_t workspace_byte_limit_;
+    string prefer_;
+};
+
+struct CudnnConvHandle{
+    cudnnTensorDescriptor_t x_desc_ ;
+    cudnnTensorDescriptor_t y_desc_ ;
+    cudnnTensorDescriptor_t bias_desc_ ;
+    cudnnFilterDescriptor_t filter_desc_ ;
+    cudnnConvolutionDescriptor_t conv_desc_ ;
```
nudles: why not make CudnnConvHandle a subclass of ConvHandle?

src/model/convolution_functions.h
```diff
@@ -0,0 +1,67 @@
...
+    Tensor workspace_;
+
+    size_t height_;
+    size_t width_;
+    size_t conv_height_;
+    size_t conv_width_;
+    size_t batchsize;
+};
+
+ConvHandle SetupConv(
```
nudles: SetupConv should also init Cudnn?

python/singa/autograd.py
xuewanqi: as in function CudnnConvForward(x,w,b,ch,cch), there needs a arg CTensor b. it is easy to do so if don't want to modified c++ codes. b isn't involved in any calculation if not bias.

python/singa/autograd.py
xuewanqi: yes, it can.

src/model/convolution_functions.cc
xuewanqi: if merged, i consider there are to many parameters in one handle. and since input tensor is needed to init cudnnhandle, so the setup can only be done after __call__(), then there is nothing to do in __init__(). 

I think ConvHandle is needed by both cpu and gpu parts, it can be set firstly in __init__(). then create a structure CpuConvHandle(), after __call__(), the program decide whether to create a GpuConvHandle(CudnnConvHandle) or CpuConvHandle.

src/model/convolution_functions.cc
```diff
@@ -0,0 +1,457 @@
...
+
+    CHECK(channels_ == ch.channels_)<<"the number of input channels mismatched.";
+
+    conv_height_ = 1;
+    if (ch.stride_h_ > 0)
+        conv_height_ = (height_ + 2 * ch.pad_h_ - ch.kernel_h_) / ch.stride_h_ + 1;
+    conv_width_ = (width_ + 2 * ch.pad_w_ - ch.kernel_w_) / ch.stride_w_ + 1;
+
+    col_height_ = ch.channels_ * ch.kernel_w_ * ch.kernel_h_;
+    col_width_ = conv_height_ * conv_width_;
```
nudles: the code here is redundant to Conv_GPU.

src/model/convolution_functions.cc
```diff
@@ -0,0 +1,457 @@
...
+
+        col_height_,
+        col_width_
+    };
+};
+
+Convolution C;
+
+Tensor CpuConvForward(const Tensor &x, Tensor &W,  Tensor &b,
+                        const ConvHandle ch, const CpuConvHandle cch){
```
nudles: it is difficult to differentiate ConvHandle and CpuConvHandle from the name.
either merge them or use a different name.

python/singa/autograd.py
nudles: why not merge __call__ into forward?

python/singa/autograd.py
```diff
@@ -583,6 +583,107 @@ def backward(self, dy):
...
+
+    	self.W.to_device(self.dev)
+    	xs = [x, self.W]
+    	
+    	self.b.to_device(self.dev)
+    	xs.append(self.b)
+    	return self._do_forward(*xs)[0]
+
+    def forward(self, *xs):
+        return singa.CudnnConvForward(xs[0], xs[1], xs[2], self.convhandle, self.cudnnconvhandle)
```
nudles: do init in forward()?

python/singa/autograd.py
```diff
@@ -583,6 +583,107 @@ def backward(self, dy):
 def flatten(x):
     return Flatten()(x)[0]
 
+class Conv2d_GPU(Operation):
```
nudles: there is no Conv2d_CPU..

python/singa/autograd.py
xuewanqi: in the design, __call__(args) do some prepare works for _do_forward(). in _do_forward(), it mainly realize two  effects: 1. record topology(except forward()), 2. call forward() function to do calculation.

src/model/convolution_functions.cc
xuewanqi: ok.i will change them.

src/model/convolution_functions.cc
xuewanqi: i will refine it.

python/singa/autograd.py
xuewanqi: i will implement it later or merge cpu and gpu part.

python/singa/autograd.py
xuewanqi: ok.

src/model/operation/convolution.cc
```diff
@@ -347,7 +347,7 @@ Tensor GpuConvBackwardW(const Tensor &dy, const Tensor &x, const Tensor &W, cons
   Tensor dW;
   dW.ResetLike(W);
 
-  dy.device()->Exec([dW, dy, x, W, cch](Context * ctx) {
+  dy.device()->Exec([dW, dy, x, &cch](Context * ctx) {
```
nudles: better use reference & for all args in [ ]

xuewanqi: ok

src/model/operation/convolution.h
```diff
@@ -0,0 +1,93 @@
...
+
+class ConvHandle {
+
+ public:
+  ConvHandle(const Tensor &input, const std::vector<size_t>& kernel_size,
+             const std::vector<size_t>& stride, const std::vector<size_t>& padding,
+             const size_t in_channels, const size_t out_channels,
+             const bool bias);
+ 
+  size_t kernel_w;
```
nudles: when the pysinga creates the convhandle, it is using int (not size_t), did you encounter any error like?
```
  File "mnist_cnn.py", line 130, in <module>
    loss, y = forward(inputs, targets)
  File "mnist_cnn.py", line 113, in forward
    y = conv1(x)
  File "/home/wangwei/miniconda3/lib/python3.6/site-packages/singa/autograd.py", line 745, in __call__
    self.inner_params['workspace_MB_limit'] * 1024 * 1024, self.inner_params['cudnn_prefer'])
  File "/home/wangwei/miniconda3/lib/python3.6/site-packages/singa/singa_wrap.py", line 2043, in __init__
    this = _singa_wrap.new_CudnnConvHandle(*args)
NotImplementedError: Wrong number or type of arguments for overloaded function 'new_CudnnConvHandle'.
  Possible C/C++ prototypes are:
    singa::CudnnConvHandle::CudnnConvHandle(singa::Tensor const &,std::vector< size_t,std::allocator< size_t > > const &,std::vector< size_t,std::allocator< size_t > > const &,std::vector< size_t,std::allocator< size_t > > const &,size_t const,size_t const,bool const,size_t const,std::string const &)
    singa::CudnnConvHandle::CudnnConvHandle(singa::Tensor const &,std::vector< size_t,std::allocator< size_t > > const &,std::vector< size_t,std::allocator< size_t > > const &,std::vector< size_t,std::allocator< size_t > > const &,size_t const,size_t const,bool const,size_t const)
    singa::CudnnConvHandle::CudnnConvHandle(singa::Tensor const &,std::vector< size_t,std::allocator< size_t > > const &,std::vector< size_t,std::allocator< size_t > > const &,std::vector< size_t,std::allocator< size_t > > const &,size_t const,size_t const,bool const)

```

xuewanqi: i didn't encounter this error. i use python 2.7 to do the test. maybe because the difference of python version?

nudles: can you try python3?
```
cmake -USE_PYTHON3=ON -USE_CUDA=ON ..
```

xuewanqi: ok. i will try and report to you.

PR #395, opened 2018-07-06
Memory Management branch: pool and swap

junzhezhang:
memory pool: full version
swap: swap synchronous and asynchronous switched off, swap schedule switched off.


nudles:
Delete all .DS_Store files.

PR #396, opened 2018-07-10
SINGA-378 Implement maxpooling operation and its related functions for autograd

xuewanqi:
To simply the design of maxpooling operation, some related functions need to be built. When maxpooling operation works, it no longer need to call corresponding layer but call these functions.

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -359,10 +359,13 @@ class SoftMaxCrossEntropy(Operation):
     def forward(self, x, t):
         self.p = singa.SoftMax(x)
         self.t = t
-        return singa.CrossEntropyFwd(self.p, t)
+        loss = CTensor((1,), self.p.device())
+        ret = singa.CrossEntropyFwd(self.p, t)
+        loss.SetFloatValue(singa.SumAsFloat(ret) / x.shape()[0])
+        return loss
 
     def backward(self, dy=1.0):
-        return singa.SoftmaxCrossEntropyBwd(self.p, self.t)
+        return singa.SoftmaxCrossEntropyBwd(self.p, self.t), None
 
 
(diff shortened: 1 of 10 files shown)


python/singa/autograd.py
```diff
@@ -758,11 +731,83 @@ def __call__(self, x):
...
+                                                    self.padding, self.in_channels, self.out_channels, self.bias,
+                                                    self.inner_params['workspace_MB_limit'] * 1024 * 1024, self.inner_params['cudnn_prefer'])
+
         self.handle.device_id = x.device.id()
 
         y = conv2d(x, self.W, self.b, self.handle)
         return y
+
+
+class MaxPool2D(Layer):
```
nudles: implement a Pool2D layer and then implment MaxPool2D and AvgPool2D?

src/model/operation/pooling.cc
```diff
@@ -0,0 +1,126 @@
...
+  method = pooling_method;
+  CHECK(method == "MAX" || method == "AVERAGE")
+      << "Padding implemented only for average and max pooling.";
+}
+
+#ifdef USE_CUDNN
+
+CudnnPoolingHandle::CudnnPoolingHandle(const Tensor &input, const std::vector<size_t>& kernel_size,
+                                       const std::vector<size_t>& stride, const std::vector<size_t>& padding,
+                                       const bool ceil_mode, const std::string pooling_method, const bool NaN_prop)
```
nudles: since you have the pooling_method (for max or avg), you can implement AvgPool2D Layer in the same way.

python/singa/autograd.py
xuewanqi: just need to change the argument form "MAX" to "AVERAGE" when initial handle,, then MaxPool2D layer changes to AvgPool2D layer

PR #397, opened 2018-07-11
SINGA-379 Implement batchnorm operation and its related functions for autograd

xuewanqi:


python/singa/autograd.py
```diff
@@ -749,3 +757,89 @@ def __call__(self, x):
...
+
+
+class _BatchNorm2d(Operation):
+
+    def __init__(self, handle):
+        self.handle = handle
+
+    def forward(self, x, scale, bias):
+        if training:
+            resultmean = CTensor([scale.shape(0)])
```
nudles: if you create a new resultemean and resultvariance, how do you accumulate the mean and var?

python/singa/autograd.py
```diff
@@ -749,3 +757,89 @@ def __call__(self, x):
...
+            resultvariance = CTensor([scale.shape(0)])
+            self.cache = (x, resultmean, resultvariance, scale)
+
+            if self.handle.device_id == -1:
+                raise NotImplementedError
+            else:
+                resultmean.ToDevice(x.device())
+                resultvariance.ToDevice(x.device())
+                return singa.GpuBatchNormForwardTraining(x, scale, bias, self.cache, self.handle)
```
nudles: what should be put in cache? what should be put in handle?

src/model/operation/batchnorm.cc
```diff
@@ -0,0 +1,177 @@
...
+Tensor GpuBatchNormForwardTraining(const Tensor& x, const Tensor& bnScale, const Tensor& bnBias,
+                                   const std::vector<Tensor>& cache, const CudnnBatchNormHandle &cbnh) {
+  CHECK_EQ(x.device()->lang(), kCuda);
+  CHECK_EQ(bnScale.device()->lang(), kCuda);
+  CHECK_EQ(bnBias.device()->lang(), kCuda);
+  CHECK_EQ(cbnh.runningMean.device()->lang(), kCuda);
+  CHECK_EQ(cbnh.runningVariance.device()->lang(), kCuda);
+  CHECK_EQ(cache[1].device()->lang(), kCuda);  //resultmean
+  CHECK_EQ(cache[2].device()->lang(), kCuda);  //resultvariance
+
```
nudles: what's the difference between runningMean and resultmean?

src/model/operation/batchnorm.h
```diff
@@ -0,0 +1,62 @@
...
+  BatchNormHandle(const float momentum, const Tensor& input, const Tensor& RunningMean, const Tensor& RunningVariance);
+
+  float factor;
+
+  size_t batchsize;
+  size_t channels;
+  size_t height;
+  size_t width;
+
+  Tensor runningMean;
```
nudles: put runningMean and runningVariance at python level?

src/model/operation/batchnorm.cc
xuewanqi: runningMean is a statistics inference of mean of whole dataset used for training. resultMean is the exact mean for the current batch of data. resultMean and resultVariance are needed when do backward, if not stored, they needs to be calculated again in backward(). resultMean is also used to estimate(calculate) runningMean, the equation is runningmean = momentum*runningmean + (1-momentum)* resultmean.

python/singa/autograd.py
xuewanqi: cache=(x,resultMean,resultVariance, scale) and handle has (runningMeaning, runningVariance).

src/model/operation/batchnorm.h
xuewanqi: runningMean and runningVariance should be in the same level with scale and bias. the only difference of them is that, runningMean and runningVariance don't need update. their date keep changing when do training. need to let batchnorm handle know the exist of them just like batchnorm layer know exist of them. their should not only be calculated gradually in training process, but also should support set initial value directly. export them to python helps pass init value from layer to handle.

python/singa/autograd.py
xuewanqi: in my former design, i let cache=(x,). then want to create resultmean and resultvariance in Forward() and push_back them to cache during the forward process. then the parameter declaration for cache cannot be set as 'const'.  then when i test, i sent GpuBatchNormForwardTraining(x, scale, bias, [x], handle), there is a argument type error. [x] cannot be understanded as vector.

PR #409, opened 2018-08-20
SINGA-383 Add Separable Convolution for autograd

xuewanqi:
The former Conv2d layer connot support Grouped Convolution, which make it difficult to realize some networks like inception net, xception net.

In this ticket, i will let Conv2d layer support Grouped convolution. and implement separable_conv2d which is the basic module in xception net by calling conv2d.

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -781,14 +781,14 @@ class SeparableConv2d(Layer):
 
     def __init__(self, in_channels, out_channels, kernel_size, stride=1, padding=0, bias=False):
 
-        self.mapping_spacial_conv = Conv2d(
+        self.spacial_conv = Conv2d(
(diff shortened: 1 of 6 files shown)


python/singa/autograd.py
```diff
@@ -743,25 +755,43 @@ def __call__(self, x):
...
 
         y = conv2d(self.handle, x, self.W, self.b)
         return y
 
 
+class SeparableConv2d(Layer):
+
+    def __init__(self, in_channels, out_channels, kernel_size, stride=1, padding=0, bias=False):
+
+        self.mapping_spacial_conv = Conv2d(
```
nudles: rename mapping_spacial_conv to spatial_conv?

xuewanqi: ok.

PR #416, opened 2018-11-17
singa-onnx

ShichengChen:
Hi,
I write some code for singa to read onnx model and convert singa computation graph to onnx model.
examples/autograd/mlp.py is about converting singa computation graph to onnx model
examples/autograd/mplonnx2singa.py is about converting onnx model to singa computation graph.
examples/autograd/onnx_model.py/ is some tool code for above two files.
python/singa/autograd.py, add some code for reading different layers' parameters and converting onnx model to singa computation graph


examples/autograd/singa-onnx/onnx_model.py
```diff
@@ -0,0 +1,43 @@
+import pickle
```
nudles: pls add the license header to every file you added

https://github.com/apache/incubator-singa/blob/master/examples/autograd/mlp.py#L1

examples/autograd/singa-onnx/mlp.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+sgd = optimizer.SGD(0.00)
+
+# training process
+for i in range(1):
+    #print('auto grad x', tensor.to_numpy(Tensor(data=inputs.data, device=inputs.data.device)))
+    x = autograd.matmul(inputs, w0)
+    x = autograd.add_bias(x, b0)
+    # x = autograd.relu(x)
```
nudles: no relu?

examples/autograd/singa-onnx/mplonnx2singa.py
```diff
@@ -0,0 +1,75 @@
+from singa import tensor
+from singa.tensor import Tensor
+from singa import autograd
+from singa import optimizer
+import numpy as np
+import onnx
+from onnx import numpy_helper
+from onnx_model import *
```
nudles: onnx and onxx_model are two python packages?

examples/autograd/singa-onnx/mplonnx2singa.py
```diff
@@ -0,0 +1,75 @@
...
+    #print('auto grad x', tensor.to_numpy(Tensor(data=inputs.data, device=inputs.data.device)))
+    loss = onnx_loss(a,model,target)
+    gradient = autograd.backward(loss)
+    for p, gp in gradient:
+        #print(p.shape)
+        #print(gp.shape)
+        gp.reshape(p.shape)
+        #print()
+        #gp = gp.reshape(p.shape)
+        #print(gp.shape)
```
nudles: pls remove the unnecessary comments.

examples/autograd/singa-onnx/onnx_model.py
```diff
@@ -0,0 +1,43 @@
...
+from singa import tensor
+from singa.tensor import Tensor
+from singa import autograd
+from singa import optimizer
+import numpy as np
+import onnx
+from onnx import numpy_helper
+
+def load_onnx_model(name = 'singonnx.pkl'):
```
nudles: should the two functions be put into autograd.py, i.e., included in singa py code?

python/singa/autograd.py
```diff
@@ -283,7 +439,8 @@ def forward(self, x):
         Returns:
             a new CTensor whose element y = x if x >= 0; otherwise 0;
         '''
-        if training:
+        self.param={'name':'LeakyRelu','x':x,'alpha':0.0}
```
nudles: 1. indent alignment
2. why do you need to add an attribute `param`?

python/singa/autograd.py
```diff
@@ -77,6 +82,156 @@ def gradients(y, dy=None):
     return grads
 
 
+
+def get_onnx_model(y, dy=None):
+    ######################
+    ##onnx
```
nudles: can you add some comments to this function?
it's difficult to understand the workflow without any description.

examples/autograd/singa-onnx/onnx_model.py
ShichengChen: yes

examples/autograd/singa-onnx/mlp.py
ShichengChen: add relu now

examples/autograd/singa-onnx/mplonnx2singa.py
ShichengChen: I merge them to single file sonnx.py(short for singa-onnx)
onnx_model is my utils file, I delete it now.
onnx is official library

examples/autograd/singa-onnx/mplonnx2singa.py
ShichengChen: yes

examples/autograd/singa-onnx/onnx_model.py
ShichengChen: yes, I moved the load_onnx_model to sonnx.py(short for singa-onnx)
the load_onnx_model function will not confuse with autograd functions

python/singa/autograd.py
ShichengChen: 1. I need to save names of models' layer
2. and the data used for the current layer.

python/singa/autograd.py
ShichengChen: Yes, I add some comments.
the main part of the function is autograd.backward.
find each node of the computational graph and save to onnx model

python/singa/sonnx.py
```diff
@@ -0,0 +1,242 @@
...
+        onnx model
+    '''
+    ######################
+    import onnx
+    from onnx import helper
+    from onnx import AttributeProto, TensorProto, GraphProto
+    from onnx import numpy_helper
+    import numpy as np
+    from singa import tensor
+    X = helper.make_tensor_value_info('X', TensorProto.FLOAT, [1, 2])
```
nudles: what does the hardcode number [1, 2] mean?

python/singa/sonnx.py
```diff
@@ -0,0 +1,242 @@
...
+        pstrd = str(op.src[0][0]) + str(dependency[op.src[0][0]])
+        cstr = str(op)
+        pstr = str(op.src[0][0])
+        #get father node's name and children nodes' names
+
+        if (pre == 'Dummy'): pstr = pre = 'X'
+        # if it is father node is dummy, the node will be input x
+        if (op.param['name'] == 'LeakyRelu'):
+        # save autograd layers param name, and use here for saving onnx model
+            node = [onnx.helper.make_node('LeakyRelu', inputs=[pstr], outputs=[cstr], )] + node
```
nudles: can create a dictionary to replace the if else check?
Otherwise, everytime a new Op is added, we have to change the code here.

python/singa/sonnx.py
```diff
@@ -0,0 +1,242 @@
...
+
+    # ready is a queue of (operation, dy list)
+    ready = deque([(y.creator, (dy,))])
+    # using bfs iterative method to check all computational graph nodes
+    not_ready = {}  # mapping: op->[dy]
+    gradients = {}  # mapping: x->dx if x.stores_grad
+    if y.stores_grad:
+        gradients[y] = dy
+
+    while len(ready) > 0:
```
xuewanqi: this block of code can be simplified.  no need to do unnecessary backward calculations and store the calculated gradients.

python/singa/sonnx.py
```diff
@@ -0,0 +1,242 @@
...
+        Args:
+            name:name of onnx model file
+        Return:
+            onnx model
+    '''
+    with open(name, 'rb') as f:
+        model = pickle.load(f)
+    return model
+
+def onnx_model_init(inputs,model):
```
xuewanqi: if can merge the function 'load_onnx_model' and 'onnx_model_init'? the first function seems unnecessary.

python/singa/sonnx.py
```diff
@@ -0,0 +1,242 @@
...
+    '''
+    a = {}
+    a['X'] = inputs
+    for i in model.graph.node:
+        if (i.op_type == 'Constant'):
+            a[str(i.output[0])] = tensor.from_numpy(numpy_helper.to_array(i.attribute[0].t))
+            a[str(i.output[0])].stores_grad = True
+    return a
+
+def onnx_loss(a,model,target):
```
xuewanqi: even the first three function can merged. just a funtion, onnx_loss(inputs, targets, model_path), to return loss tensor?

python/singa/sonnx.py
```diff
@@ -0,0 +1,242 @@
...
+    '''
+    get onnx model loss
+    Args:
+        a: singa computational graph nodes dictionary
+        model: onnx model
+        target:
+    Return:
+        loss for onnx model
+    '''
+    for i in model.graph.node:
```
xuewanqi: if there is any improvement? not only for the loading, but also for the generation of nodes.

python/singa/sonnx.py
ShichengChen: It's the input size.
I just checked that if the input size is fit for the model structure, the result is right.
This number can be ignored.

python/singa/sonnx.py
ShichengChen: yes

python/singa/sonnx.py
ShichengChen: yes

python/singa/sonnx.py
ShichengChen: yes

python/singa/sonnx.py
ShichengChen: onnx_init load model weights but onnx_loss load other parts.
So I want to separate them.
Is that ok?

python/singa/sonnx.py
ShichengChen: onnx has a python api for generate nodes.

examples/singa-onnx/mlp.py
```diff
@@ -0,0 +1,104 @@
...
+
+from singa import tensor
+from singa.tensor import Tensor
+from singa import autograd
+from singa import optimizer
+from singa import *
+from singa import sonnx
+
+import numpy as np
+#import caffe2.python.onnx.backend as backend
```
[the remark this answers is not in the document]
ShichengChen: Thank you for your comments, I just fixed these problems.

examples/onnx/mlponnx2singa.py
```diff
@@ -0,0 +1,72 @@
...
+# specific language governing permissions and limitations
+# under the License.
+#
+
+from singa import tensor
+from singa.tensor import Tensor
+from singa import autograd
+from singa import optimizer
+import numpy as np
+from singa import sonnx
```
moazreyad: When a deep learning system like SINGA supports ONNX, usually there are two parts to implement: the first is Exporting to ONNX (frontend) and the second is Importing ONNX models (backend). (See https://github.com/onnx/tutorials).

Please consider separating the sonnx.py file into two python files/modules, one for backend and one for frontend.

examples/onnx/mlp.py
```diff
@@ -0,0 +1,102 @@
...
+    for p, gp in gradient:
+        sgd.apply(0, gp, p, '')
+    if (i % 100 == 0):
+        print('training loss = ', tensor.to_numpy(loss)[0])
+
+
+model=sonnx.get_onnx_model(loss)
+
+with open('singonnx.pkl', 'wb') as output:
+    pickle.dump(model,output)
```
moazreyad: ONNX models should not be saved with pickle. Please use ONNX python API : `onnx.save(onnx_model, 'path/to/the/model.onnx')` instead. See [here](https://github.com/onnx/onnx/blob/master/docs/PythonAPIOverview.md#saving-an-onnx-model).

python/singa/sonnx.py
```diff
@@ -0,0 +1,204 @@
...
+    load onnx model graph and load weights
+    input:
+    input data and file name of onnx model
+
+    return:
+     a graph node dictionary
+     model: graph model
+    '''
+    f = open(name, 'rb')
+    model = pickle.load(f)
```
moazreyad: ONNX models should not be loaded with pickle. Please use ONNX python API :` onnx_model = onnx.load('path/to/the/model.onnx')` instead. See [here](https://github.com/onnx/onnx/blob/master/docs/PythonAPIOverview.md#loading-an-onnx-model).

examples/onnx/mlp.py
```diff
@@ -0,0 +1,102 @@
...
+
```
moazreyad: This example file is not complete because it does not show how the ONNX model that was exported from SINGA can be imported into another AI library. I would suggest to complete this example by loading the ONNX model from the file to another deep learning software such as MxNet, PyTorch or TensorFlow.

examples/onnx/mlponnx2singa.py
```diff
@@ -0,0 +1,72 @@
...
+
+label = to_categorical(label, 3).astype(np.float32)
+print('train_data_shape:', data.shape)
+print('train_label_shape:', label.shape)
+
+inputs = Tensor(data=data)
+target = Tensor(data=label)
+
+
+a,model = sonnx.onnx_model_init(inputs,'singonnx.pkl')
```
moazreyad: This example is not complete because it does not show how the ONNX model that was exported from another AI library can be imported into SINGA. I would suggest to complete this example by creating a model in another deep learning software such as MxNet, PyTorch or TensorFlow. Then exporting this model using ONNX and loading the onnx model from the file to SINGA.

examples/onnx/mlp.py
nudles: I think we just need to add another example to show that SINGA can load the onnx file exported by another library, e.g., the [file exported by pytorch](https://pytorch.org/tutorials/advanced/ONNXLive.html#convert-the-pytorch-models-to-onnx-models). If we add the example to show that the exported onnx file from SINGA can be imported into pytorch, we have to make pytorch a dependent lib of SINGA.

python/singa/sonnx.py
ShichengChen: yes, I changed to onnx.save and onnx.load

examples/onnx/mlp.py
ShichengChen: yes, I supported the pytorch's onnx model

examples/onnx/mlponnx2singa.py
nudles: We can have a link to the Pytorch/MxNet code for exporting the model via onnx. in our repo, we only include the code for loading the model via onnx. Then the example is complete.

nudles:
@moazreyad do you have other comments?

python/singa/autograd.py
```diff
@@ -586,6 +586,52 @@ def device_check(self, *inputs):
             if var.device.id() != x_device:
                 var.to_device(x_device)
 
+    def find_sublayers(self):
+        # return a list whose elements are in form of (attribute_name,
+        # sublayer)
```
nudles: you need to 
1. remove these code from #422 
2. rebase your branch onto the latest master branch. then you will have these code in your branch

python/singa/sonnx.py
```diff
@@ -0,0 +1,270 @@
...
+#from .tensor import einsum
+from autograd import *
+from singa.tensor import to_numpy
+from autograd import _Conv2d,_Pooling2d,_BatchNorm2d
+
+def onnx_model_init(path):
+    '''
+    load onnx model graph and load weights
+    input:
+    input data and file name of onnx model
```
nudles: the input is a single argument..

python/singa/sonnx.py
```diff
@@ -0,0 +1,270 @@
...
+            elif(i.op_type == 'Concat'):
+                oper[str(i.output[0])] = autograd.cat((oper[str(i.input[0])], oper[str(i.input[1])]),int(i.attribute[0].i))
+            elif(i.op_type == 'Tanh'):
+                oper[str(i.output[0])] = autograd.tanh(oper[str(i.input[0])])
+            elif (i.op_type == 'Sigmoid'):
+                oper[str(i.output[0])] = autograd.sigmoid(oper[str(i.input[0])])
+            elif (i.op_type == 'Mul'):
+                oper[str(i.output[0])] = autograd.mul(oper[str(i.input[0])],oper[str(i.input[1])])
+            elif (i.op_type == 'MaxPool'):
+                oper[str(i.output[0])] = layer[str(i.output[0])](oper[str(i.input[0])])
```
nudles: since the MaxPool, Conv, BatchNorm have already been combined and converted into layer, can we replace the conditional check, i.e., `elif` by using the layer directly from `layer` dictionary.

python/singa/sonnx.py
ShichengChen: yes

python/singa/sonnx.py
ShichengChen: yes

examples/onnx/cnn.py
```diff
@@ -0,0 +1,146 @@
...
+        for i in range(batch_number):
+            inputs = tensor.Tensor(device=dev, data=x_train[
+                                   i * 100:(1 + i) * 100], stores_grad=False)
+            targets = tensor.Tensor(device=dev, data=y_train[
+                                    i * 100:(1 + i) * 100], requires_grad=False, stores_grad=False)
+
+            loss, y = forward(inputs, targets)
+
+            model = sonnx.get_onnx_model(loss, inputs, targets)
+            onnx.save(model, 'cnn.onnx')
```
nudles: why do you save the model in every iteration?

examples/onnx/cnn.py
```diff
@@ -0,0 +1,146 @@
...
+    for epoch in range(epochs):
+        for i in range(batch_number):
+            inputs = tensor.Tensor(device=dev, data=x_train[
+                                   i * 100:(1 + i) * 100], stores_grad=False)
+            targets = tensor.Tensor(device=dev, data=y_train[
+                                    i * 100:(1 + i) * 100], requires_grad=False, stores_grad=False)
+
+            loss, y = forward(inputs, targets)
+
+            model = sonnx.get_onnx_model(loss, inputs, targets)
```
nudles: rename it to to_onnx_model() ?

examples/onnx/mplonnx2singa.py
```diff
@@ -0,0 +1,69 @@
...
+
+label = to_categorical(label, 3).astype(np.float32)
+print('train_data_shape:', data.shape)
+print('train_label_shape:', label.shape)
+
+inputs = Tensor(data=data)
+target = Tensor(data=label)
+
+
+model = sonnx.from_onnx_model('linear.onnx')
```
nudles: the from_onnx_model and get_onnx_model are inconsistent..
get_onnx_model() has args: loss, input and target. 
but the model created using from_onnx_model only has args: input.

I think the get_onnx_model should accept two args: [input list], [output_list]
and the model constructed using from_onnx_model should take [input_list] as input and generates [output_list]

examples/onnx/cnn.py
ShichengChen: yes

examples/onnx/cnn.py
ShichengChen: i took the function out from loop

examples/onnx/mplonnx2singa.py
ShichengChen: yes

moazreyad:
Currently onnx is not installed with SINGA by default. So if the user tries to run the code, he will get an error:

> ImportError: No module named onnx

We need to add onnx to SINGA python dependencies [here](https://github.com/apache/incubator-singa/blob/master/python/setup.py.in).

nudles:
Agree. @ShichengChen pls update the dependency in setup.in and [meta.yml](https://github.com/apache/incubator-singa/blob/master/tool/conda/singa/meta.yaml) files.

examples/onnx/linear.py
```diff
@@ -0,0 +1,89 @@
...
+    gradient = autograd.backward(loss)
+    for p, gp in gradient:
+        sgd.apply(0, gp, p, '')
+    if (i % 100 == 0):
+        print('training loss = ', tensor.to_numpy(loss)[0])
+
+
+model=sonnx.ONNXm.to_onnx_model([loss],[inputs,target])
+
+onnx.save(model, 'linear.onnx')
```
moazreyad: This example creates a file 'linear.onnx' successfully. However, the model in this file is not correct.

If you run the following code:

```
import onnx
model = onnx.load("linear.onnx")
onnx.checker.check_model(model)
```

There is an error: 

> ValidationError: No Schema registered for CrossEntropy with domain_version of 8

I think CrossEntropy is not one of the [onnx operators](https://github.com/onnx/onnx/blob/master/docs/Operators.md). My understanding is that cross entropy is used only while training the model and it is not saved with the model in the onnx file. 

May be you can save _x3_ instead of _loss_? (assuming _x3_ is the output of the model)

examples/onnx/pytorchMPL.py
```diff
@@ -0,0 +1,25 @@
+from scipy import misc
```
moazreyad:
```suggestion
```
this import is not used in the file

examples/onnx/mplonnx2singa.py
```diff
@@ -0,0 +1,69 @@
+#
```
moazreyad: File name is mplonnx2singa.py. but I don't know what is mpl.

Did you mean MLP instead? so the file name should be mplonnx2singa.py.

examples/onnx/mlp.py
```diff
@@ -0,0 +1,102 @@
...
+    gradient = autograd.backward(loss)
+    for p, gp in gradient:
+        sgd.apply(0, gp, p, '')
+    if (i % 100 == 0):
+        print('training loss = ', tensor.to_numpy(loss)[0])
+
+
+model=sonnx.ONNXm.to_onnx_model([loss],[inputs,target])
+
+onnx.save(model, 'mlp.onnx')
```
moazreyad: Same problem here for cross entropy. Please see this [comment](https://github.com/apache/incubator-singa/pull/416/files#r248686478).

examples/onnx/mplonnx2singa.py
```diff
@@ -0,0 +1,69 @@
...
+
+inputs = Tensor(data=data)
+target = Tensor(data=label)
+
+
+model = sonnx.from_onnx_model('mlp.onnx')
+print('finish init')
+sgd = optimizer.SGD(0.00)
+
+# training process
```
moazreyad: Why there is a training process here? The model in the file mlp.onnx was already trained in mlp.py. 

It is expected to use the trained model here by passing an input and predicting the output without training again.

examples/onnx/mnist_cnn.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+        accuracy_rate = accuracy(tensor.to_numpy(y),
+                                 tensor.to_numpy(targets))
+        print('accuracy is:', accuracy_rate, 'loss is:',tensor.to_numpy(loss)[0])
+        for p, gp in autograd.backward(loss):
+            sgd.update(p, gp)
+
+model = sonnx.ONNXm.to_onnx_model([loss], [inputs,targets])
+onnx.save(model, 'cnn.onnx')
```
moazreyad: Please check the generated model. see this [comment](https://github.com/apache/incubator-singa/pull/416/files#r248686478). May be you can save _y_ instead of _loss_ if _y_ is the model output.

examples/onnx/cnnonnx2singa.py
```diff
@@ -0,0 +1,123 @@
...
+    # operations initialization
+    conv1 = autograd.Conv2d(1, 32, 3, padding=1)
+    conv21 = autograd.Conv2d(32, 16, 3, padding=1)
+    conv22 = autograd.Conv2d(32, 16, 3, padding=1)
+    linear = autograd.Linear(32 * 28 * 28, 10)
+
+    model = sonnx.from_onnx_model('cnn.onnx')
+    print('finish init')
+    autograd.training = True
+    # training process
```
moazreyad: Why there is another training process here? The model in the file `cnn.onnx` was already trained in `mnist_cnn.py`.

It is expected to use the trained model here by passing an input image and predicting the output without training again.

python/singa/sonnx.py
```diff
@@ -0,0 +1,299 @@
...
+from . import layer
+from singa.proto import model_pb2
+from . import singa_wrap as singa
+#from .tensor import einsum
+from autograd import *
+from singa.tensor import to_numpy
+from autograd import _Conv2d,_Pooling2d,_BatchNorm2d
+
+
+class ONNXm(Layer):
```
moazreyad: I think this class is performing both Importing ONNX models (backend) and Exporting to ONNX (frontend). It will be better to [separate the importing (backend) from the exporting (frontend)](https://github.com/apache/incubator-singa/pull/416#discussion_r236562502). 

Now the frontend function is a static method INSIDE the ONNXm class (see from_onnx_model() ), while the backend function is a function OUTSIDE the ONNXm class (see to_onnx_model() ).

This is not a problem, but the class which supports the backend operations must inherit from from `onnx.backend.base.Backend`. So there are two options: 

1. make ONNXm inherits from onnx.backend.base.Backend:

```
from onnx.backend.base import Backend
class ONNXm(backend)
...
```
in this case the ONNXm class will do both backend and frontend together, so the backend functions must be implemented INSIDE it.

or (better)

2. create another new backend class which calls this one as needed. The new backend class must also inherit from `onnx.backend.base.Backend`:

```
from onnx.backend.base import Backend
class SingaBackend(backend)
...
```

See [CNTK backend](https://github.com/onnx/onnx-cntk/blob/765f51f5b26f62933bb2e29c29661a152c001b89/onnx_cntk/backend.py#L13) and [TensforFlow backend](https://github.com/onnx/onnx-tensorflow/blob/c4a75b09e85ffb7b5afda14b64194ca972f957bf/onnx_tf/backend.py#L32) .

nudles:
> 
> 
> There are examples with this pull request, but examples can not replace unit tests. At least the backend should include testing. In order to [test the backend](https://github.com/onnx/onnx/blob/master/docs/OnnxBackendTest.md), we need to call the function `onnx.backend.test.BackendTest` which expects a class that inherits from `onnx.backend.base.Backend`. See for example, [CNTK backend test](https://github.com/onnx/onnx-cntk/blob/765f51f5b26f62933bb2e29c29661a152c001b89/onnx_cntk/tests/onnx_backend_test.py#L20) and [TensforFlow backend test](https://github.com/onnx/onnx-tensorflow/blob/c4a75b09e85ffb7b5afda14b64194ca972f957bf/test/backend/test_onnx_backend.py#L18).
> 
> We also need to have a page in the documentation with the list of supported operations for import and export. See for example [MxNet ONNX Operator Coverage](https://cwiki.apache.org/confluence/display/MXNET/ONNX+Operator+Coverage).

I think the current backend API in onnx is more proper for declarative libraries like TF and CNTK, and for inference only. It tries to wrap everything (the whole training process including data loading, loss computing, sgd, etc.) into a BackendRep's run function. Then we cannot get the gradients of parameters for model update.


I suggest to implement the frontend and backend via the following two functions, which can also be used to implement the onnx backend.

```
singa_model=from_onxx_model(onnx_model)  # the singa_model can be used as singa_model(input) that generates output
onnx_model = to_onnx_model(singa_model, input)  # can refer to pytorch's frontend.
```

examples/onnx/cnnonnx2singa.py
ShichengChen: I think that we can provide one more options to continue to train the model such as fine tune?

python/singa/sonnx.py
ShichengChen: I separated frontend and backend. backend include one-time backend and multiple times backend.
should I keep cross entropy layer?

ShichengChen: I separated frontend and backend. backend include one-time backend and multiple times backend.
I just removed cross entropy layer

examples/onnx/cnnonnx2singa.py
nudles: we can load onnx model directly for inference.
we can also load it to continue the training.

include/singa/core/device.h
```diff
@@ -49,123 +49,123 @@ namespace singa {
 /// Allocate memory and execute Tensor operations.
 /// There are three types of devices distinguished by their programming
```
nudles: I think your editor changed this file automatically.
pls revert it back.

src/core/device/device.cc
```diff
@@ -19,58 +19,58 @@
 #include "singa/core/device.h"
 
 namespace singa {
-Device::Device(int id, int num_executors)
-    : id_(id), num_executors_(num_executors) {
-  // TODO(wangwei) create scheduler and vm.
-  host_ = defaultDevice;
-}
+    Device::Device(int id, int num_executors)
```
nudles: revert this file.

src/model/operation/convolution.cc
```diff
@@ -218,7 +221,7 @@ CudnnConvHandle::CudnnConvHandle(const Tensor &input,
 #endif
                                              ));
   if (CUDNN_MAJOR >= 7 && groups > 1) {
-    CUDNN_CHECK(cudnnSetConvolutionGroupCount(conv_desc, groups));
+    //CUDNN_CHECK(cudnnSetConvolutionGroupCount(conv_desc, groups));
```
nudles: why comment it out?

ShichengChen: I used the wrong cudnn version before. I have uncommented it

src/core/device/device.cc
ShichengChen: yes

include/singa/core/device.h
ShichengChen: yes

examples/onnx/cnnonnx2singa.py
moazreyad: @nudles Yes, but I do not see an example for loading the model for inference and I am worried that the model parameters are not saved and loaded correctly. This is why I propose to show that inference is working without re-training.

python/singa/sonnx.py
nudles: According to [onnx backend implementation](https://github.com/onnx/onnx/blob/master/docs/ImplementingAnOnnxBackend.md), we need to implement two backends.
one for one time inference; one for running the onnx model times (but just do the conversion for only one time).

python/singa/sonnx.py
```diff
@@ -0,0 +1,289 @@
...
+            elif(i.op_type == 'Concat'):
+                oper[str(i.output[0])] = autograd.cat((oper[str(i.input[0])], oper[str(i.input[1])]),int(i.attribute[0].i))
+            elif(i.op_type == 'Tanh'):
+                oper[str(i.output[0])] = autograd.tanh(oper[str(i.input[0])])
+            elif (i.op_type == 'Sigmoid'):
+                oper[str(i.output[0])] = autograd.sigmoid(oper[str(i.input[0])])
+            elif (i.op_type == 'Mul'):
+                oper[str(i.output[0])] = autograd.mul(oper[str(i.input[0])],oper[str(i.input[1])])
+            elif (i.op_type == 'Conv'):
+                handledic = Backend.convhandle(i.name,handledic,oper[str(i.input[0])],model)
```
nudles: if the handle is used immediately, do you need to put it into a dict?
creating the handle in every iteration is expensive!

ShichengChen: I had written a if check :   
` if(name in handledic):return handledic`
if I create handle before, will not create handle again

python/singa/sonnx.py
```diff
@@ -18,272 +18,359 @@
...
+            if node.name not in handles:
+                handles[node.name] = handle(inputs, attrs)
+            tensors[outputs[0]] = forward(handles[node.name], *inputs)
+        elif op == 'Concat':
+            tensors[outputs[0]] = autograd.cat(tuple(inputs), attrs['axis'])
+        else:
+            raise NotImplementedError('Not supported op: {}'.format(op))
+
+    def run(self, input):
+        # input_dict: dict from input name to numpy array
```
nudles: input has singa tensor or numpy array?

python/singa/sonnx.py
```diff
@@ -18,272 +18,359 @@
...
+                name2tensor[node.output[0]] = t
+
+        return SingaBackendRep(model, device, name2tensor)
+
+    @classmethod
+    def run_node(cls, node, inputs, device, outputs_info=None, **kwargs):
+        '''
+        Args:
+            node: onnx node proto
+            inputs: dictionary of name to numpy array; the names should match
```
nudles: numpy array or singa tensor?

python/singa/sonnx.py
nudles: if singa tensor, then input can be a list of tensors (each tensor's name field must be initialized).

python/singa/sonnx.py
```diff
@@ -18,272 +18,359 @@
...
-
-
-
+                    ready.append(srcop)
+
+    for x in inputsx:
+        dtype = TensorProto.FLOAT
+        if y.dtype == tensor.int32:
+            dtype = TensorProto.INT
+        X.append(helper.make_tensor_value_info(inputs[0] , dtype, x.shape))
```
nudles: inputs[0] is a str or a tensor?

python/singa/sonnx.py
```diff
@@ -18,272 +18,359 @@
...
-                                             value=numpy_helper.from_array(dummy1))] + node
+                node.append(helper.make_node('AveragePool',
+                                             inputs=inputs,
+                                             outputs=outputs,
+                                             name=op.name,
+                                             kernel_shape=k,
+                                             pads=p,
+                                             strides=s))
+        elif (isinstance(op, autograd._BatchNorm2d)):
+            dummy0 = tensor.to_numpy(tensor.Tensor(device=op.running_mean.device(), data=op.running_mean))
```
nudles: dummy tensors will be processed in line 355. no need to process them here explicitly.

python/singa/sonnx.py
```diff
@@ -18,272 +18,359 @@
...
+        output_name(y.creator, 0), TensorProto.FLOAT, y.shape)]
+
     while len(ready) > 0:
         op = ready.pop()
-        if isinstance(op, Dummy):continue
+        assert not isinstance(op, autograd.Dummy)
+        outputs = [output_name(op, idx)
+                   for _, idx in op.y_id2idx.items()]
+        inputs = [output_name(srcop, srcop.y_id2idx[yid]) for (srcop, yid, _, _) in op.src]
         curop = str(op).split('.')[-1].split(' ')[0]
```
nudles: where is curop used?

python/singa/autograd.py
```diff
@@ -1115,7 +1161,8 @@ def __init__(self, kernel_size, stride=None, padding=0):
 
 
 class Tanh(Operation):
-
+    def __init__(self):
```
nudles: if the init function does nothing, then no need to define it.

examples/onnx/cnnonnx2singa.py
```diff
@@ -105,19 +105,13 @@ def accuracy(pred, target):
...
 
     model = onnx.load('cnn.onnx')
-    rep = sonnx.BackendRep(model,dev)
-    #####backend run multiple times
+    rep = sonnx.prepare(model, dev)
     print('finish init')
     autograd.training = True
     # training process
     for epoch in range(1):
         inputs = tensor.Tensor(device=dev, data=x_train[0:100], stores_grad=False)
```
nudles: input tensor should have a name.

examples/onnx/cnnonnx2singa.py
```diff
@@ -105,19 +105,13 @@ def accuracy(pred, target):
...
         targets = tensor.Tensor(device=dev, data=y_train[0:100], requires_grad=False, stores_grad=False)
```
nudles: target also need a name

python/singa/sonnx.py
```diff
@@ -18,272 +18,359 @@
...
+            tensors[outputs[0]] = autograd.cat(tuple(inputs), attrs['axis'])
+        else:
+            raise NotImplementedError('Not supported op: {}'.format(op))
+
+    def run(self, input):
+        # input_dict: dict from input name to numpy array
+        tensors = self.tensor_dict.copy()
+        key=self.model.graph.input[0].name
+        oname = self.model.graph.output[0].name
+        tensors[key] = input[0]
```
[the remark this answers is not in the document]
ShichengChen: I used name system to replace hard code

python/singa/sonnx.py
```diff
@@ -18,272 +18,359 @@
...
+    @classmethod
+    def prepare(cls,
+                model,  # type: ModelProto
+                device,  # type: Text
+                **kwargs  # type: Any
+                ):  # type: (...) -> Optional[BackendRep]
+        '''
+        Args:
+            model: onnx model proto
+            device: 'CPU' or 'GPU'
```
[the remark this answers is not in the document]
ShichengChen: singa device and i update the comments

python/singa/sonnx.py
ShichengChen: singa tensor

python/singa/sonnx.py
ShichengChen: singa tensor

python/singa/sonnx.py
ShichengChen: a tensor

python/singa/sonnx.py
ShichengChen: i just rechecked the code, running var and running mean will not processed in line 355
so i write some extra code here

python/singa/sonnx.py
ShichengChen: name of operator

python/singa/autograd.py
ShichengChen: to get names from base class operator, we need to use __init__()

examples/onnx/cnnonnx2singa.py
ShichengChen: if input order here is the same as input order in to_onnx_model, default name system works

python/singa/tensor.py
```diff
@@ -87,10 +87,10 @@ class Tensor(object):
                      grad but not store grad; But if a tensor stores grad
                      then it must require grad.
     '''
-
+    tensor_count = 0
```
nudles: where is this var used?

python/singa/sonnx.py
```diff
@@ -325,7 +328,7 @@ def output_name(op, index):
                                          name=op.name,
                                          momentum=0.9))
```
nudles: don't hard code!

python/singa/sonnx.py
nudles: running var and running mean are tensors generated by Dummy. why are they missed in line 355?

python/singa/sonnx.py
```diff
@@ -257,24 +255,28 @@ def to_onnx_model(inputs, y, model_name='sonnx'):
...
+        X.append(helper.make_tensor_value_info(x.name, dtype, x.shape))
     Y = [helper.make_tensor_value_info(
-        output_name(y.creator, 0), TensorProto.FLOAT, y.shape)]
+        output_name(y.creator), TensorProto.FLOAT, y.shape)]
 
     while len(ready) > 0:
         op = ready.pop()
         assert not isinstance(op, autograd.Dummy)
-        outputs = [output_name(op, idx)
+        outputs = [output_name(op)
```
nudles: if there are multiple outputs, then they all would have the same name.

ShichengChen: I added a if code

python/singa/sonnx.py
ShichengChen: yes

python/singa/sonnx.py
ShichengChen: #[(<singa.autograd.Sigmoid object at 0x7fd5ec09cb90>, 140556764852432, None, False),
# (<singa.autograd.Dummy object at 0x7fd5ec09c390>, 140556764824208,
# <singa.tensor.Tensor object at 0x7fd5ec09c290>, True),
 # (<singa.autograd.Dummy object at 0x7fd5ec09c490>, 140556764824528,
 # <singa.tensor.Tensor object at 0x7fd5ec09c3d0>, True),
 # (<singa.autograd.Dummy object at 0x7fd5ec09c590>, 140556764824784, None, False),
 # (<singa.autograd.Dummy object at 0x7fd5ec09c690>, 140556764825040, None, False)])
# two dummy operators do not have values, so take the values from handle

python/singa/tensor.py
ShichengChen: `if name is None:
            self.name = 'Dummy#{}'.format(Tensor.tensor_count)
            Tensor.tensor_count += 1
        else:
            self.name = name`
add default name, we did not have self.name before, so i added it

ShichengChen: I imitated from autograd.py's operator name code

src/api/model_operation.i
```diff
@@ -74,12 +74,16 @@ class CudnnConvHandle: public ConvHandle {
                   const std::string& prefer = "fastest");
   bool bias_term;
   size_t batchsize;
-    size_t pad_w;
+  size_t pad_w;
   size_t pad_h;
   size_t stride_h;
   size_t stride_w;
   size_t padding_h;
```
nudles: pad_w, padding_w? redundant?

src/api/model_operation.i
```diff
@@ -118,6 +123,14 @@ class CudnnPoolingHandle : public PoolingHandle {
 
   int pooled_height;
   int pooled_width;
+  int kernel_h;
```
nudles: these are inherited from PoolingHandle?

python/singa/sonnx.py
```diff
@@ -319,14 +320,21 @@ def output_name(op):
                                              pads=p,
                                              strides=s))
         elif (isinstance(op, autograd._BatchNorm2d)):
-            print(op,op.src)
+            #[(<singa.autograd.Sigmoid object at 0x7fd5ec09cb90>, 140556764852432, None, False),
+            # (<singa.autograd.Dummy object at 0x7fd5ec09c390>, 140556764824208,
+            # <singa.tensor.Tensor object at 0x7fd5ec09c290>, True),
```
nudles: what are these??

python/singa/sonnx.py
```diff
@@ -258,8 +258,8 @@ def to_onnx_model(inputs, y, model_name='sonnx'):
     dependency = autograd.infer_dependency(y.creator)
     ready = deque([y.creator])
 
-    def output_name(op):
-        return '{}'.format(op.name)
+    def output_name(op,extra=''):
+        return '{}'.format(op.name+str(extra))
```
nudles: extra is a str? then why do you the type conversion?

src/api/model_operation.i
ShichengChen: yes, i removed redundant variables

src/api/model_operation.i
ShichengChen: yes

python/singa/sonnx.py
ShichengChen: default is string, but if pass int to the function, it can convert to string automatically

python/singa/sonnx.py
ShichengChen: singa operation information.
include output variables and input variables.
I want to show that some of the dummy variables do not have tensor values

python/singa/autograd.py
```diff
@@ -614,10 +682,28 @@ def __call__(self, x):
             y = add_bias(y, self.b, axis=0)
         return y
 
+    def get_params(self):
+        if self.bias:
+            return {'W': self.W, 'b': self.b}
+        else:
+            return {'W': self.W}
+
+    def set_params(self, **parameters):
```
nudles: will this function be called? or set_one_param() is enough?

python/singa/tensor.py
```diff
@@ -107,9 +107,14 @@ def __init__(self, shape=(), device=None, dtype=float32,
         self.dtype = self.data.data_type()
         self.requires_grad = requires_grad
         self.stores_grad = stores_grad
+        if name is None:
+            self.name = 'Dummy#{}'.format(Tensor.tensor_count)
```
nudles: we only need to give a name to the input tensors.
for the tensors generated by operations, the name should be set by the output_name() function.

python/singa/autograd.py
ShichengChen: no, I will not use this function

python/singa/tensor.py
ShichengChen: `node.append(helper.make_node('Concat',
                                         inputs=inputs,
                                         outputs=outputs,
                                         name=op.name,
                                         axis=op.axis))`
for the tensors generated by operations
output parameter is set by output_name() 
name parameter is set by op.name
should i change name parameter to output_name() too? same as output parameter?

moazreyad:
It is very hard to follow this pull request. There are more than 130 comment. I think the singa-onnx task is too large to be implemented in one pull request. Also the files in this pull request were changed several times and it is not easy to follow the changes.

Can we split the task into smaller sub tasks, so it can be easier to follow?

May be also we can merge some of the completed work in this pull request. We may merge with the master branch if the changes are good enough, or in another temporary branch for the onnx task. Merging will allow others to create other pull requests to add, fix or discuss changes instead of putting everything in one pull request.

nudles:
+1. 
@ShichengChen pls close this PR and create a new PR using the code at https://github.com/nudles/incubator-singa/tree/pr416-new1. Let me rebase it first.

PR #423, opened 2018-11-27
SINGA-411 Local variable 'params' referenced before assignment

dbxinj:


python/singa/net.py
```diff
@@ -472,7 +472,7 @@ def save(self, f, buffer_size=10, use_pickle=False):
                 f = f[0:-4]
             sp = snapshot.Snapshot(f, True, buffer_size)
             v = tensor.from_numpy(np.array([__version__]))
-            params['SINGA_VERSION'] = v
+            sp.write('SINGA_VERSION', __version__)
```
nudles: `__version__` should be `v`.

PR #424, opened 2018-11-29
SINGA-411 Local variable 'params' referenced before assignment

xuewanqi:
- fixed bug.

### DIFF
--- a/python/singa/net.py
+++ b/python/singa/net.py
@@ -472,7 +472,7 @@ class FeedForwardNet(object):
                 f = f[0:-4]
             sp = snapshot.Snapshot(f, True, buffer_size)
             #v = tensor.from_numpy(np.array([__version__]))
-            #sp.write(b'SINGA_VERSION', v)
+            #sp.write('SINGA_VERSION', v)
             for (name, val) in zip(self.param_names(), self.param_values()):
                 val.to_host()
                 sp.write(name, val)
--- a/python/singa/net.py
+++ b/python/singa/net.py
@@ -471,8 +471,8 @@ class FeedForwardNet(object):
             if f.endswith('.bin'):
                 f = f[0:-4]
             sp = snapshot.Snapshot(f, True, buffer_size)
-            v = tensor.from_numpy(np.array([__version__]))
-            sp.write('SINGA_VERSION', v)
+            #v = tensor.from_numpy(np.array([__version__]))
+            #sp.write(b'SINGA_VERSION', v)
             for (name, val) in zip(self.param_names(), self.param_values()):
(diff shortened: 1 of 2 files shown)


python/singa/net.py
```diff
@@ -471,8 +471,8 @@ def save(self, f, buffer_size=10, use_pickle=False):
             if f.endswith('.bin'):
                 f = f[0:-4]
             sp = snapshot.Snapshot(f, True, buffer_size)
-            v = tensor.from_numpy(np.array([__version__]))
-            sp.write('SINGA_VERSION', v)
+            #v = tensor.from_numpy(np.array([__version__]))
```
nudles: the two lines should not be commented out

src/io/snapshot.cc
```diff
@@ -48,7 +48,7 @@ Snapshot::Snapshot(const std::string& prefix, Mode mode, int max_param_size /*in
     text_reader_ptr->Open(prefix + ".desc");
     std::string key, val;
     while (text_reader_ptr->Read(&key, &val)) {
-      if (key == "SINGA_VERSION")
+      if (key == "0")
```
nudles: this is not the best to way fix the bug.
need to update the textfilereader and writer: https://github.com/apache/incubator-singa/blob/master/src/io/textfile_writer.cc#L52

python/singa/net.py
xuewanqi: maybe just write version information to .desc file is sufficient. if use sp.write('singa_version'，v) to write, then the useful information just be stored in .bin file. It brings unnecessary complexity to decode. And for the corresponding .desc file, it should be 'parameter name: SINGA_VERSION	data type: 0	dim: 1	shape: 1', which are all the useless information.

src/io/snapshot.cc
xuewanqi: can figure out a way to retain if (key == "SINGA_VERSION")

xuewanqi: yes. it's a little bit strange. the code can pass in my local environment. 
i am trying to find the reason

python/singa/net.py
xuewanqi: I rethink about it. Store all useful information in one particular file may be better, which mean store all information just in .bin file. this is the file computer to read. it requires to write v tensor. 
computer no need to read .desc file, it can be designed just for user to read. 
In former implementation, computer read .desc file just to do version check, which can be simplified.

nudles: it's fine to remove the version check. the desc file is human readable, which is convenient for checking the meta info of the .bin file.

PR #426, opened 2018-12-02
SINGA-405 Add roles to contributors

moazreyad:
This pull request is an invitation for adding the roles for each contributor in the SINGA team in order to satisfy the [Apache Maturity Model](https://community.apache.org/apache-way/apache-project-maturity-model.html):

CS10: The project maintains a public list of its contributors who have decision power -- the project's PMC (Project Management Committee) consists of those contributors.

See [this example.](http://pulsar.apache.org/team/)

Please complete the _roles_ column in this PR with missing values, because I do not have the roles of everyone.

This will also improve the voting process in SINGA because some voting results may not be accurate according to the Apache process, like [this one](http://mail-archives.apache.org/mod_mbox/singa-dev/201811.mbox/%3CCAH%3DWUUb79%2Bvob%2BcBZMXAc4u4R6C6PHyjRrckPnctdhpNUggEcQ%40mail.gmail.com%3E), because the list of contributors with voting power was not made public.


nudles:
should PMC be PPMC (the podling PMC)?

moazreyad:
Yes, PPMC is used before graduation and PMC is used after graduation.

During the graduation process itself, the graduation guidelines says:

> CS10: The project maintains a public list of its contributors who have decision power -- the project's **PMC** (Project Management Committee) consists of those contributors.

So my understanding was that the role 'PMC' cab be used during the graduation. I also copied 'PMC' from Pulsar team page example. 

However, to be accurate, we should use PPMC now. Then after graduation, we should change it to PMC. This what Pulsar [did](https://github.com/apache/pulsar/commit/828e22df135b17379f12681a4ccad68eaaa0948a#diff-c4e00d34eaffca83e6173fce77e0c629).

moazreyad:
I think we need to review and merge this pull request before graduation.
This is related to CS10 in [the maturity model](https://cwiki.apache.org/confluence/display/SINGA/Maturity+model+assessment).

PR #431, opened 2019-01-01
Integrated MKL-dnn for operation in model 

dcslin:
- Integrated MKL-dnn for operations below fwd/bwd w/ tests
  - Conv2D
  - BatchNorm 4D/2D
  - Pooling Max/Avg
- Updated CMake for flag: USE_MKLDNN


src/model/operation/batchnorm.cc
```diff
@@ -17,8 +17,187 @@ BatchNormHandle::BatchNormHandle(const float momentum, const Tensor& input) {
...
+    y.ResetLike(x);
+
+
+    Tensor w = get_bn_weight_from(bnScale, bnBias);
+
+    y.device()->Exec([&y, &x, &w, &bnh](Context *ctx) {
+      try {
+        auto eng = *ctx->engine;
+        using namespace mkldnn;
+        auto x_mem = memory({{{bnh.x_dims}, bnh.dtype, bnh.data_memory_format}, eng}, x.block()->mutable_data());
```
nudles: can we use x.block()->data() instead of mutable_data()?
data() returns a const pointer, i.e., the pointed memory can NOT be updated.
mutable_data() returns a raw pointer, i.e., the pointed memory can be updated.

src/model/operation/batchnorm.cc
```diff
@@ -17,8 +17,187 @@ BatchNormHandle::BatchNormHandle(const float momentum, const Tensor& input) {
...
+    delete (dx_md);
+    delete (bn_fwd_d);
+    delete (bn_fwd_pd);
+#endif // USE_MKLDNN
+  }
+
+#ifdef USE_MKLDNN
+
+  Tensor CpuBatchNormForwardInference(const BatchNormHandle &bnh, const Tensor& x, const Tensor& bnScale, const Tensor& bnBias,
+                                                        Tensor& running_mean, Tensor& running_var){
```
nudles: running_mean and running_var are not used?

src/model/operation/convolution.cc
```diff
@@ -99,6 +175,42 @@ Tensor CpuConvBackwardx(const Tensor &dy, Tensor &W, const Tensor &x,
...
+  dx.ResetLike(x);
+
+  dy.device()->Exec([&x, &dx, &dy, &W, &ch](Context *ctx) {
+    Block *wblock = W.block(), *dyblock = dy.block(), *dxblock = dx.block(), *inblock = x.block();
+
+    try {
+      auto eng = *ctx->engine;
+      using namespace mkldnn;
+      auto x_mem = memory({{{ch.x_dims}, ch.dtype, memory::format::nchw}, eng}, inblock->mutable_data());
+      auto w_mem = memory({{{ch.w_dims}, ch.dtype, memory::format::goihw}, eng}, wblock->mutable_data());
```
nudles: is the group convolution supported?

CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ OPTION(USE_OPENCL "Use OpenCL" OFF)
 OPTION(ENABLE_DIST "Enable distributed training" OFF)
 OPTION(DISABLE_WARNINGS "Disable warnings under windows" ON)
 OPTION(USE_MODULES "Compile dependent libs as submodules together with singa" OFF)
+OPTION(USE_MKLDNN "Use mkl-dnn libs" ON)
```
moazreyad: This line will make USE_MKLDNN enabled by default. Is this the required default configuration?

Also please remember to add this new option to the[ installation documentation](https://github.com/apache/incubator-singa/blob/master/doc/en/docs/installation.md).

src/model/operation/batchnorm.cc
dcslin: to feed the data into mkldnn, `memory primitive` is required. and constructor only accept `void * data_handle`. and it seems that this is the only way for now. https://github.com/intel/mkl-dnn/issues/387

src/model/operation/convolution.cc
dcslin: yes. mkldnn support "group"

src/model/operation/batchnorm.cc
dcslin: fixed and updated

src/model/operation/batchnorm.cc
nudles: then keep mutable_data(). copying the data is expensive.

nudles:
I think the error from travis is because of the dependency not meet.
Profobuf requires higher version of gcc/gxx. 
You can try to update the [this file](https://github.com/apache/incubator-singa/blob/master/tool/conda/singa/meta.yaml) with

```
build:
        - gcc_impl_linux-64 7.3.0 habb00fd_1
        - gcc_linux-64 7.3.0 h553295d_3
        - gxx_impl_linux-64 7.3.0 hdf63c60_1
        - gxx_linux-64 7.3.0 h553295d_3
        - libgcc-ng 8.2.0 hdf63c60_1
        - libstdcxx-ng 8.2.0 hdf63c60_1
run:
        - libgcc-ng >=7.3.0
        - libstdcxx-ng >=7.3.0
```

CMakeLists.txt
dcslin: Hi Moaz, Thank you for the advice, this is resolved.

PR #438, opened 2019-03-07
*fix bug: segmentation fault at test case

joddiy:
*fix bug:
    - allow reshaping on an empty tensor then give it data will invoke segmentation fault.
*how to fix:
    - when reshape, add a check requiring the original shape's volume must be same with new one.
    - fix corresponding test case(test_mse, test_cross_entropy, test_tensor_math)

### DIFF
--- a/src/core/tensor/tensor.cc
+++ b/src/core/tensor/tensor.cc
@@ -1228,6 +1228,8 @@ void SoftmaxCrossEntropyBwd(const Tensor &t, Tensor *p) {
 // if tensor is not transposed yet, we change the shape and generate new strides
 // if tensor is already transposed, we reallocate the memory and generate strides
 Tensor& Tensor::Reshape(const Shape &shape) {
+  // Check original shape's volume is same with the new one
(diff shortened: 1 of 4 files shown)


nudles:
The tranpose operation should return a new tensor that shares the data with this tensor if possible; otherwise return a new tensor that copies the data.

joddiy:
The transpose is another problem, I think yisen will submit a new PR(otherwise I'll simbit a new PR to copy that).

nudles:
maybe the current transpose implementation is ok. T() and Tensor::Transpose() change itself. The global transpose return a view if possible.  No need to change now.

joddiy:
But the current transpose implementation cannot pass the test case. The test case
`
  Tensor o = t.T();
  EXPECT_EQ(t.shape()[0],  o.shape()[1]);
  EXPECT_EQ(t.shape()[1],  o.shape()[0]);
`
seems the transpose func needs to return a new transposed tensor, rather than change itself.

PR #457, opened 2019-05-04
test pipeline, why not returning 1 on python test error??????

3dprintscanner:


3dprintscanner:
Further to this, it appears that on Ubuntu 18.04, the conda config for building the library pulls the 3.x.x version of libprotobuf, changing the build to go against the 2.x.x protobuf version causes this to break from a broken linked dependency in _singa_wrap.so

moazreyad:
This PR is required for enabling the python test with continuous integration. Without this PR, the user must check the error messages of the python test manually by looking at the log and the python test errors are not shown as check errors in Travis CI or Github Actions.

[Here](https://github.com/moazreyad/singa/runs/886716641?check_suite_focus=true) is the Gihub action with python errors without this PR (the check passes even if there are errors) and [here](https://github.com/moazreyad/singa/runs/886745085?check_suite_focus=true) is the Github action with this PR (the check fails correctly due to errors in python test).

PR #462, opened 2019-05-21
SINGA-333 Add ONNX backend test

moazreyad:


nudles:
@joddiy  can we merge this one? or we need to wait until the operators are all implemented?

joddiy:
I think we cannot directly use all the onnx test cases. Since, first, it's hard to integrate the sonnx into it. Second, there are lots of test cases we cannot use because we don't have some functionalities, such as dilation...
So far, I've written a new test file to adapt the test cases from onnx, so that we can discard some cases that we cannot implement.

joddiy:
please refer to [here](https://github.com/apache/incubator-singa/pull/534)

moazreyad:
But if we can't run onnx backend test cases, then we simply can't claim that we support onnx backend.

We don't have to pass all the test cases, but at least some of them should pass. Currently [all of the test cases fail](https://issues.apache.org/jira/projects/SINGA/issues/SINGA-458), which can be understood as if the current support of onnx backend is 0%.

joddiy:
> But if we can't run onnx backend test cases, then we simply can't claim that we support onnx backend.
> 
> We don't have to pass all the test cases, but at least some of them should pass. Currently [all of the test cases fail](https://issues.apache.org/jira/projects/SINGA/issues/SINGA-458), which can be understood as if the current support of onnx backend is 0%.

Hi, moazreyad, I know what you mean. Please check the link [here](https://github.com/apache/incubator-singa/pull/534). I mean you cannot just easily call all the onnx test cases, it cannot run. I have adapted the onnx backend test cases from onnx there, and discard the cases that we cannot support.

PR #475, opened 2019-07-03
SINGA-463 fixed python test_operation

dcslin:


python/singa/autograd.py
```diff
@@ -924,17 +924,14 @@ def __init__(
...
-            self.b = Tensor(shape=b_shape, requires_grad=True, stores_grad=True)
-            self.b.set_value(0.0)
-        else:
-            # to keep consistency when to do forward.
-            self.b = None
-            # Tensor(data=CTensor([]), requires_grad=False, stores_grad=False)
+        b_shape = (self.out_channels,)
+        # forward method required bias variable to proceed (dummy b tensor)
+        self.b = Tensor(shape=b_shape, requires_grad=self.bias, stores_grad=self.bias)
+        self.b.set_value(0.0)
```
nudles: sometimes, the model is configured without using bias.

dcslin: Hi, yes. This change is to cater the backend cpp API, which requires bias to be passed in.
for example: `src/model/operation/convolution.h`
```c++
Tensor GpuConvForward(const Tensor &x, const Tensor &W, const Tensor &b, const CudnnConvHandle &cch);
``` 
In the frontend, python side, it will not be updated:
`python/singa/autograd.py`
```python
if self.handle.bias_term:
    # compute db
```

nudles: the bias tensor cannot be created here because onnx will consider it a real bias (onnx has no bias_term flag).
[this line](https://github.com/apache/incubator-singa/blob/b99aaa9f87ccbd4cf111e76013b8387b7129bd8b/python/singa/autograd.py#L791) should also be updated to set b as an optional argument `forward(self, x, W, b=None)`.
You can create the tensor for b before calling the CPP function.

python/singa/autograd.py
```diff
@@ -924,17 +924,14 @@ def __init__(
...
-            b_shape = (self.out_channels,)
-            self.b = Tensor(shape=b_shape, requires_grad=True, stores_grad=True)
-            self.b.set_value(0.0)
-        else:
-            # to keep consistency when to do forward.
-            self.b = None
-            # Tensor(data=CTensor([]), requires_grad=False, stores_grad=False)
+        b_shape = (self.out_channels,)
+        # forward method required bias variable to proceed (dummy b tensor)
+        self.b = Tensor(shape=b_shape, requires_grad=self.bias, stores_grad=self.bias)
```
nudles: maybe use an empty tensor like this https://github.com/apache/incubator-singa/blob/0aae4c12cbc94d88a6c77793a7106c292925aac7/python/singa/autograd.py#L801

nudles: It also has some updates on batchnorm operation, https://github.com/apache/incubator-singa/pull/468/files#diff-a236c495700ad62b39e28a4ece74711fR1095

PR #476, opened 2019-07-04
fixed minor bugs python test_loss tset_mkldnn

dcslin:


dcslin:
Hi, with this last fix, I can run all unittest with no error at my end.

PR #482, opened 2019-07-28
SINGA-474 hardsigmoid operator

ShichengChen:

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -372,10 +372,11 @@ class HardSigmoid(Operation):
         Returns:
             a CTensor for the result
         """
-        if training:
-            self.input = x
 
         x = singa.AddFloat(singa.MultFloat(x,self.alpha),self.gamma)
+        if training:
+            self.cache = x
+
         x = singa.ReLU(x)
         mask = singa.LTFloat(x, 1.0)
         mask2 = singa.GEFloat(x, 1.0)
@@ -391,9 +392,9 @@ class HardSigmoid(Operation):
             a (CTensor) dx
         """
 
-        x = singa.AddFloat(singa.MultFloat(self.input,self.alpha),self.gamma)
-        mask0 = singa.GTFloat(x, 0.0)
-        mask1 = singa.LTFloat(x, 1.0)
+
+        mask0 = singa.GTFloat(self.cache, 0.0)
+        mask1 = singa.LTFloat(self.cache, 1.0)
 
         mask = singa.__mul__(mask0,mask1)
         return singa.__mul__(singa.MultFloat(mask, self.alpha),dy)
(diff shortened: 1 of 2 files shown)


python/singa/autograd.py
```diff
@@ -358,6 +358,51 @@ def relu(x):
...
+        """
+        if training:
+            self.input = x
+
+        x = singa.AddFloat(singa.MultFloat(x,self.alpha),self.gamma)
+        x = singa.ReLU(x)
+        mask = singa.LTFloat(x, 1.0)
+        mask2 = singa.GEFloat(x, 1.0)
+
+        ans = singa.__add__(singa.__mul__(x, mask),mask2)
```
nudles: why add the mask2 here?

test/python/test_operation.py
```diff
@@ -321,7 +321,34 @@ def test_LeakyRelu(self):
...
+        a=0.5
+        g=0.6
+        y = np.clip(x * 0.5 + 0.6, 0, 1)
+        x = tensor.from_numpy(x)
+        x.to_device(cpu_dev)
+
+        result = autograd.hardsigmoid(x,a,g)
+        dx = result.creator.backward(x.data)
+
+        np.testing.assert_array_almost_equal(tensor.to_numpy(result), y, decimal=5)
```
nudles: backward is not tested..

python/singa/autograd.py
ShichengChen: if x >=1 then x == 1
I first made x[x>=1]=0 and then x[x>=1]+=1

test/python/test_operation.py
ShichengChen: yes, I just added test

python/singa/autograd.py
joddiy: can we write like this?
```
x = singa.AddFloat(singa.MultFloat(x, self.alpha), self.gamma)
x = singa.__clamp__(x, 0.0, 1.0)
```

python/singa/autograd.py
```diff
@@ -358,6 +358,51 @@ def relu(x):
...
+
+    def backward(self, dy):
+        """
+        Args:
+            dy (CTensor): data for the dL / dy, L is the loss
+        Returns:
+            a (CTensor) dx
+        """
+
+        x = singa.AddFloat(singa.MultFloat(self.input,self.alpha),self.gamma)
```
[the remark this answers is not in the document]
ShichengChen: yes, I just cached. We do not have clip operation now

python/singa/autograd.py
nudles: relu will turn x into x if x>=0 and 0 otherwise;
then, you do not need to add 1?

nudles:
pls resolve the conflicts

PR #483, opened 2019-07-28
SINGA-474 identity operator

ShichengChen:


test/python/test_operation.py
```diff
@@ -322,6 +322,27 @@ def test_LeakyRelu(self):
...
+    def test_Identity_cpu(self):
+        x = np.array([-0.9, -0.3, -0.1, 0.1, 0.5, 0.9]).reshape(3, 2).astype(np.float32)
+        y = x.copy()
+        x = tensor.from_numpy(x)
+        x.to_device(cpu_dev)
 
+        result = autograd.identity(x)
+        dx = result.creator.backward(x.data)
+
+        np.testing.assert_array_almost_equal(tensor.to_numpy(result), y, decimal=5)
```
joddiy: please add the test case of backward propagation.

test/python/test_operation.py
```diff
@@ -322,6 +322,27 @@ def test_LeakyRelu(self):
...
+    def test_Identity_gpu(self):
+        x = np.array([-0.9, -0.3, -0.1, 0.1, 0.5, 0.9]).reshape(3, 2).astype(np.float32)
+        y = x.copy()
+        x = tensor.from_numpy(x)
+        x.to_device(gpu_dev)
+
+        result = autograd.identity(x)
+        dx = result.creator.backward(x.data)
+
+        np.testing.assert_array_almost_equal(tensor.to_numpy(result), y, decimal=5)
```
joddiy: please add the test case of backward propagation.

test/python/test_operation.py
```diff
@@ -322,6 +336,37 @@ def test_LeakyRelu(self):
         np.testing.assert_array_almost_equal(tensor.to_numpy(result), XT)
         self.check_shape(dx.shape(), (3, 2))
 
+    def test_Identity_cpu(self):
+        x = np.array([-0.9, -0.3, -0.1, 0.1, 0.5, 0.9]).reshape(3, 2).astype(np.float32)
+        y = x.copy()
+        lossf=lambda x:np.sum(x)
```
joddiy: why the lossf is ```np.sum```? I think for the identity, the lossf is just ```x```.

ShichengChen: The derivative of np.sum is one, so I use sum as loss function. 
The eval_gradient function only accepts loss function, so I made loss function here.

joddiy: Got it.  It's ok.

joddiy:
ready for merge

PR #488, opened 2019-07-29
SINGA -475 add Sign operator to singa

pinpom:


joddiy:
ready for merge

python/singa/autograd.py
```diff
@@ -1828,3 +1828,18 @@ def backward(self, dy):
...
+            self.input = a
+        return singa.Sign(a)
+
+    def backward(self, dy):
+        dx = singa.MultFloat(dy, 0.0)
+        return dx
+
+
+def sign(a):
+    return Sign()(a)[0]
```
chrishkchris: change a to x
def sign(x):
    return Sign()(x)[0]

python/singa/autograd.py
```diff
@@ -1828,3 +1828,18 @@ def backward(self, dy):
 
 def leakyrelu(x, a=0.01):
     return LeakyRelu(a)(x)[0]
+
+
+class Sign(Operation):
+    def forward(self, a):
```
chrishkchris: Add the default initializer:

    def __init__(self):
        super(Sign, self).__init__()

python/singa/autograd.py
pinpom: input of the function is "a"

python/singa/autograd.py
pinpom: fix the code

test/python/test_operation.py
```diff
@@ -610,6 +610,42 @@ def test_Atanh_gpu(self):
...
+        DY = np.ones((3, 2), dtype = np.float32)
+
+        x = tensor.from_numpy(X)
+        dy = tensor.from_numpy(DY)
+        x.to_device(gpu_dev)
+        dy.to_device(gpu_dev)
+
+        result = autograd.sign(x)
+        dx = result.creator.backward(dy.data)
+        #dx = [x/|x|]'
```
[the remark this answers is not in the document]
pinpom: fix the code

chrishkchris:
Thanks! Ready for merge

python/singa/autograd.py
chrishkchris: Thanks for the modification. The below is the reason autograd.py this is needed.

This function passes the name "Sign" to the superclass (Operation) initialize.
The superclass Operation initialize is as follows:
```
class Operation(object):
    op_count = 0
    def __init__(self, name=None):
        if name is None:
            self.name = "{}#{}".format(
                self.__class__.__name__, Operation.op_count
            )
            Operation.op_count += 1
        else:
            self.name = name
```
If there is no initialzer in the subclass, the superclass initializer will be called with the "name=None"

python/singa/autograd.py
```diff
@@ -1828,3 +1828,21 @@ def backward(self, dy):
...
+    def __init__(self):
+        super(Sign, self).__init__()
+
+    def forward(self, a):
+        if training:
+            self.input = a
+        return singa.Sign(a)
+
+    def backward(self, dy):
+        dx = singa.MultFloat(dy, 0.0)
```
nudles: the dx is always 0?

chrishkchris:
In my opinion:
For the function y=sign(x) , y = 1 when x > 0, y = -1 when x < 0. Therefore, the derivative is 0 except x = 0. 
The function y=sign(x) is discontinuous at x=0 where the derivative is undefined in the mathematical logic. However, an defined derivative leads to a nan number, so we used a derivative of zero also at x=0.

As a result, dx is always 0 while the output shape (size of array) is the same as input shape.

chrishkchris:
This is what tensorflow use:
```cpp
Status SignGrad(const Scope& scope, const Operation& op,
                const std::vector<Output>& grad_inputs,
                std::vector<Output>* grad_outputs) {
  auto shape = Shape(scope, op.input(0));
  auto zero = Cast(scope, Const(scope, 0.0), op.input(0).type());
  auto dx = Fill(scope, shape, zero);
  grad_outputs->push_back(dx);
  return scope.status();
}
REGISTER_GRADIENT_OP("Sign", SignGrad);
```
seems to fill the dx with all zero

chrishkchris:
there are conflicts and there are two solutions (same as what happened in log operator): 
1. merge this branch with the latest master
2. checkout a new branch from the latest master. Then copy the added code into the autograd.py and test_operations.py; send a new PR.

PR #492, opened 2019-07-29
Make singa use multiple memory pools

ShichengChen:


include/singa/core/device.h
```diff
@@ -295,7 +295,9 @@ class Platform {
   /// Create a set of CudaGPU Device using given GPU IDs.
   static const std::vector<std::shared_ptr<Device>>
   CreateCudaGPUsOn(const std::vector<int> &devices, size_t init_size = 0);
-  
+
+    static std::vector<std::shared_ptr<Device> > allRet;
```
nudles: what is allRet and retUsed?
pls add some comments.

nudles: should they be static members of Platform?

src/core/device/platform.cc
```diff
@@ -120,6 +121,26 @@ Platform::CreateCudaGPUs(const size_t num_devices, size_t init_size) {
 
 const vector<shared_ptr<Device>>
 Platform::CreateCudaGPUsOn(const vector<int> &devices, size_t init_size) {
+
```
nudles: it would be better to add a lock here.
if two threads call this function at the same time, this function's output is then not deterministic.

src/core/device/platform.cc
```diff
@@ -133,8 +154,11 @@ Platform::CreateCudaGPUsOn(const vector<int> &devices, size_t init_size) {
   vector<shared_ptr<Device> > ret;
   for (auto device : devices) {
     auto dev = std::make_shared<CudaGPU>(device, pool);
-    ret.push_back(dev);
+    UsedDevice[device]=dev;
   }
+  int count = Platform::GetNumGPUs();
+  for(int i = 0;i < count;i++)if(UsedDevice[i]!=nullptr)
```
nudles: no comments and the code is not formatted following google style.

src/core/memory/memory.cc
```diff
@@ -39,12 +39,10 @@ CnMemPool::CnMemPool(int numDevices, size_t init_size, size_t max_size) {
     conf_.add_device(i);
   conf_.set_init_size(init_size);
   conf_.set_max_size(max_size);
-  CHECK_LT(++pool_count, 2) << "CnMemPool must be used as a singleton.";
```
nudles: must it be a singleton?
can we have a pool per GPU?

PR #493, opened 2019-07-30
SINGA-473 Autograd Trigonometry: Backward Test

chrishkchris:
Test Result:

root@26c9db193eb0:~/incubator-singa/build# python3 /root/incubator-singa/test/python/test_operation.py
....................................
Ran 36 tests in 1.010s

OK

### DIFF
--- a/test/python/test_operation.py
+++ b/test/python/test_operation.py
@@ -65,6 +65,17 @@ def prepare_inputs_targets_for_rnn_test():
     targets = [t0, t1, t2]
     return inputs, targets, h0
 
+def numpy_unary_ops_backward(func, x, dy, h=0.0005):
+    # input func: numpy unary function for testing, e.g. np.cos
+    # input x: input value to the unary function (numpy array)
+    # input dy: output gradient (numpy array)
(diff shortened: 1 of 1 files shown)


test/python/test_operation.py
```diff
@@ -65,6 +65,17 @@ def prepare_inputs_targets_for_rnn_test():
     targets = [t0, t1, t2]
     return inputs, targets, h0
 
+def numpy_unary_ops_backward(func, x, dy, h=0.0005):
```
nudles: This is not correct.
You can either use this [one](http://cs231n.github.io/optimization-1/#gradcompute), or compute the gradient explicitly, e.g., gradient of cos() is sin()

chrishkchris: OK, I will change the code using one of your two methods (the link provided or compute the gradient explicitly).

The code I did was only the diagonal of the gradient matrix, which is only suitable for unary operator where one output is determined by one input, without the coupling from other inputs.

chrishkchris: Changed the code by computing the gradient explicitly, the accuracy check is up to 5 decimals.

PR #496, opened 2019-07-31
SINGA-474 Mean operator

ShichengChen:

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -328,6 +328,38 @@ class Dummy(Operation):
         return "{}_g".format(self.name)
 
 
+class Mean(Operation):
+    def __init__(self):
+        super(Mean, self).__init__()
+
+    def forward(self, *l):
+        """
+        Args:
+            l: a list of CTensor
+            element-wise mean operator
+        Returns:
+            a new CTensor
+        """
+        if training:
+            self.l = len(l)
+        assert(len(l)>0);
+        x = l[0]
+        for i in range(1,len(l)):
+            x +=l[i]
+        return singa.MultFloat(x,1/len(l))
+
+    def backward(self, dy):
+        """
+        Args:
+            dy(CTensor): dL / dy
+        Returns:
+            a list of dx(CTensor)
+        """
+        return [singa.MultFloat(dy,1/self.l)]*self.l
+
+def mean(*l):
+    return Mean()(*l)[0]
+
 class ReLU(Operation):
     def __init__(self):
         super(ReLU, self).__init__()
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -343,9 +343,10 @@ class Mean(Operation):
         if training:
             self.l = len(l)
         assert(len(l)>0);
-        x = l[0]
-        for i in range(1,len(l)):
-            x +=l[i]
+        x = singa.Tensor(list(l[0].shape()),l[0].device())
+        x.SetFloatValue(0.0)
+        for i in range(len(l)):
+            x+=l[i]
         return singa.MultFloat(x,1/len(l))
 
     def backward(self, dy):
--- a/test/python/test_operation.py
+++ b/test/python/test_operation.py
@@ -296,7 +296,46 @@ class TestPythonOperation(unittest.TestCase):
 
         np.testing.assert_array_almost_equal(tensor.to_numpy(result), XT)
         self.check_shape(dx.shape(), (3, 2))
-        
+
+
+    def test_Mean_gpu(self):
+        x0 = np.array([-0.9, -0.3, -0.1, 0.1, 0.5, 0.9]).reshape(3, 2).astype(np.float32)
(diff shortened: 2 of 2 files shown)


python/singa/autograd.py
```diff
@@ -354,6 +354,40 @@ def backward(self, dy):
         return singa.__mul__(dy, dx)
 
 
+def relu(x):
+    return ReLU()(x)[0]
+
+class Mean(Operation):
+    def __init__(self):
```
joddiy: do you need some comments to indicate this function is element-wise mean?

test/python/test_operation.py
```diff
@@ -322,6 +335,46 @@ def test_LeakyRelu(self):
...
+        x1.to_device(gpu_dev)
+
+        result = autograd.mean(x0,x1)
+        dy = tensor.from_numpy(np.ones((3,2)).astype(np.float32))
+        dy.to_device(gpu_dev)
+        dx0,dx1 = result.creator.backward(dy.data)
+
+        np.testing.assert_array_almost_equal(tensor.to_numpy(result), y, decimal=5)
+        np.testing.assert_array_almost_equal(tensor.to_numpy(tensor.from_raw_tensor(dx0)), grad, decimal=2)
+        np.testing.assert_array_almost_equal(tensor.to_numpy(tensor.from_raw_tensor(dx1)), grad, decimal=2)
```
joddiy: This grad is the gradient at x0, not x1.

test/python/test_operation.py
```diff
@@ -37,7 +37,20 @@
 def _tuple_to_string(t):
     lt = [str(x) for x in t]
     return '(' + ', '.join(lt) + ')'
-
+def eval_numerical_gradient(f, x,y):
```
joddiy: This function calculates the gradient of f at x, not y. Please be careful when you use it.

python/singa/autograd.py
```diff
@@ -354,6 +354,40 @@ def backward(self, dy):
         return singa.__mul__(dy, dx)
 
 
+def relu(x):
+    return ReLU()(x)[0]
+
+class Mean(Operation):
```
joddiy: do you need some comments to indicate this function is element-wise mean?

test/python/test_operation.py
```diff
@@ -322,6 +335,48 @@ def test_LeakyRelu(self):
         np.testing.assert_array_almost_equal(tensor.to_numpy(result), XT)
         self.check_shape(dx.shape(), (3, 2))
 
+    def test_Mean_gpu(self):
+        x0 = np.array([-0.9, -0.3, -0.1, 0.1, 0.5, 0.9]).reshape(3, 2).astype(np.float32)
+        x1 = np.array([0, -0.3, 0, 0.1, 0, 0.9]).reshape(3, 2).astype(np.float32)
+        y = (x0+x1)/2
+        lossf =lambda x,y:np.sum((x+y)/2)
+        grad=eval_numerical_gradient(lossf,x0,x1)
+        grad1=eval_numerical_gradient(lossf,x1,x0)
```
joddiy: you can write it here because it doesn't matter when you change the order of x0 and x1. But when you write DIV function, you can not just switch it, because 'x0/x1' is different from '1x/x0'

joddiy: For example, the function is ```(X + X')/2```, at the first call, you calculate the gradient of x0 based on (x0+x1)/2, but at the second call, you calculate the gradient at x1 based on (x1+x0)/2. I think the two are still same.

joddiy: I guess you can use this function:
```
def eval_numerical_gradient_b(f, x, y, reverse = False):
    h = 0.00001
    grad = np.zeros(x.shape)
    t = y if reverse else x
    fx = f(x, y)
    it = np.nditer(t, flags=['multi_index'], op_flags=['readwrite'])
    while not it.finished:
        _it = it.multi_index
        old_value = t[_it]
        t[_it] = old_value + h # increment by h
        fth = f(x, y) # evaluate f(x + h)
        t[_it] = old_value # restore to previous value (very important!) 
        grad[_it] = (fth - fx) / h # the slope
        it.iternext() # step to next dimension
    return grad
```
if the reverse if False, this function gets the grads of x based on f(x,y), if reverse if true, it gets the grads of y still based on f(x,y)

python/singa/autograd.py
```diff
@@ -354,6 +354,41 @@ def backward(self, dy):
...
+            element-wise mean operator
+        Returns:
+            a new CTensor
+        """
+        if training:
+            self.l = len(l)
+        assert(len(l)>0);
+        x = l[0]
+        for i in range(1,len(l)):
+            x = singa.__add__(x,l[i])
```
nudles: this is not efficient. you can create an empty tensor and then accumulate each of l into it.

nudles: use inplace addition.
like x+= l[i]

ShichengChen:
```
x=tensor and set tensor to zero
for i in range(len(l)):
    x += l[i]
```
Is that what you mean?

ShichengChen: yes, I just changed the code

PR #500, opened 2019-08-08
SINGA-478 (Python 3 uses __itruediv__ instead of __idiv__)

chrishkchris:
We need to add  `__itruediv__` for python 3 in tensor.py because the original `__idiv__` is not supported by python 3 anymore.
 
To understand the problem, let's study the following code first:
```
from singa import tensor
from singa import device
import numpy as np

Y = np.ones(shape=[10],dtype=np.float32) * 10.0
y = tensor.from_numpy(Y)
y.to_device(device.get_default_device())

def divide(y):
   y /= 10

divide(y)
print(tensor.to_numpy(y))
```
 Without adding the `__itruediv__` function, the result is as follows, which means that the /= operation is not in place operation: 
`[10. 10. 10. 10. 10. 10. 10. 10. 10. 10.]`
After adding the _itruediv_ function, the result is as follows, which means that the /= operation is in place operation:
`[1. 1. 1. 1. 1. 1. 1. 1. 1. 1.]`
This is because the `__idiv__` operation is for python 2, while `__itruediv__` is for python 3. Therefore, if we do not add the `__itruediv__` operator in tensor.py, it just uses a default operation which is not in place operation.

Meanwhile, I am not sure if I also need to add `__truediv__` for python 3 to acts as  `__div__` for python 2.


nudles:
Should we need to remove `__idiv__` if it is not used by Py3?

chrishkchris:
Yes, I guess py2 is not used by us anymore for long time (e.g. the new dockerfile uses Miniconda3 py36 and py37). I have resubmitted the file, just changed `__idiv__` to `__itruediv__`

PR #504, opened 2019-08-12
SINGA-481 Reconsturct soonx

joddiy:


test/python/test_onnx.py
```diff
@@ -0,0 +1,202 @@
...
+
+from cuda_helper import gpu_dev, cpu_dev
+
+import numpy as np
+
+autograd.training = True
+
+
+class TestPythonOnnxFrontend(unittest.TestCase):
+
```
nudles: Can we use onnx's test code?
https://github.com/onnx/onnx/tree/master/onnx/backend/test/case

joddiy: These tese cases are only backend part, that is, the onnx's test case is they make two tensors, one as input, one as output, then they convert the op to other framework, run it and then transfer back the output, they check whether the framework's output is same with the pre-defined onnx's output.
They haven't tested the frontend part.
Our test cases are converting the singa op to onnx, and convert back, to check the singa ops after and before converted whether are same.

joddiy: And, please don't merge this PR now.
I found I omited a commit, and I also want to finish the transfer learning firstly.

moazreyad: > They haven't tested the frontend part.

Yes, they don't test the frontend part because it is _backend_ test. 

I think we should separate the backend test from the frontend test. The backend test must be done using the standard onnx test (onnx.backend.test.BackendTest) similar to [this pull request](https://github.com/apache/incubator-singa/pull/462). 

Otherwise, we can not claim that SINGA implements ONNX backend if the standard [ONNX backend test fails](https://issues.apache.org/jira/projects/SINGA/issues/SINGA-458). We can not ignore the standard test and create our own test that passes successfully while the standard ONNX test fails.

joddiy: Yes, I think you are right, we should also test their backend test cases. However, I'm worrying about whether we can finish it in time before this Friday. 
I guess I should focus on continuing to implement the onnx operators and transfer learning.
Also, I'm a little worrying about the autograd we implemented might have some problems so that it cannot pass the backend test cases. It will cost more time to fix it up...
So, I plan to use our test cases at first to make sure the onnx part is ok. After this Friday's demo, we can spend more time on testing those onnx backend test cases.

moazreyad: > Also, I'm a little worrying about the autograd we implemented might have some problems so that it cannot pass the backend test cases. It will cost more time to fix it up...

It can be useful to write the current problems that prevents standard ONNX backend tests from passing. Then we can discuss the alternative solutions to them. 

Because if there is a module such as autograd or any other part of the project that must be updated to enable ONNX support, may be it is better to discover this issue as early as possible before proceeding too much with an architecture or an implementation that must be changed soon.

PR #505, opened 2019-08-13
SINGA-474 sum operator

ShichengChen:
SINGA-474 sum operator for autograd and ONNX operator

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -516,6 +516,26 @@ class SoftMax(Operation):
 def softmax(x, axis=0):
     return SoftMax(axis)(x)[0]
 
+class Sum(Operation):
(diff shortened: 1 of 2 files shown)


python/singa/autograd.py
```diff
@@ -440,6 +440,21 @@ def add_bias(x, b, axis=0):
     return AddBias(axis)(x, b)[0]
 
 
+class Sum(Operation):
+    def __init__(self):
```
nudles: this is different to the definition of onnx sum operator.https://github.com/onnx/onnx/blob/master/docs/Operators.md#Sum

ShichengChen: yes, I just changed the code

PR #508, opened 2019-08-14
SINGA-474 transpose operator

ShichengChen:
SINGA-474 transpose operator for onnx operator and autograd

### DIFF
--- a/test/python/test_operation.py
+++ b/test/python/test_operation.py
@@ -991,7 +991,43 @@ class TestPythonOperation(unittest.TestCase):
 
         np.testing.assert_array_almost_equal(tensor.to_numpy(result), XT, decimal=5)
         np.testing.assert_array_almost_equal(tensor.to_numpy(tensor.from_raw_tensor(dx)), DX, decimal=5)
-    
-    
+
+
+    def test_transpose_cpu(self):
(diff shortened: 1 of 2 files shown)


test/python/test_operation.py
```diff
@@ -992,18 +992,55 @@ def test_Sqrt_gpu(self):
         np.testing.assert_array_almost_equal(tensor.to_numpy(result), XT, decimal=5)
         np.testing.assert_array_almost_equal(tensor.to_numpy(tensor.from_raw_tensor(dx)), DX, decimal=5)
 
+        
+    def test_transpose_cpu(self):
+        x = np.array([0.1,-1.0,0.4,4.0,-0.9,9.0]).reshape(3,2,1).astype(np.float32)
+        y = x.transpose(1,2,0)
+        dy = np.ones(y.shape, dtype = np.float32)
+        grad = np.ones((3,2,1), dtype = np.float32)
```
nudles: use np.random.rand

nudles: what is grad?

ShichengChen: I just changed my wrong code.
grad is dx and I used transpose for dy

PR #510, opened 2019-08-14
SINGA 475 - add min operator

pinpom:


python/singa/autograd.py
```diff
@@ -1945,3 +1945,29 @@ def backward(self, dy):
...
+        mask = singa.__add__(mask00,mask11)
+        
+        if training:
+            self.mask = mask
+            self.mask0 = mask0
+            self.mask1 = mask1
+        return mask
+
+    def backward(self, dy):
+        return (self.mask0,self.mask1)
```
nudles: dy is not multiplied in the backward?

python/singa/autograd.py
```diff
@@ -1945,3 +1945,29 @@ def backward(self, dy):
...
+    def forward(self, a, b):
+        m = singa.__sub__(a,b)
+        mask0 = singa.LTFloat(m,0) 
+        mask00 = singa.__mul__(singa.LEFloat(m,0),a)
+        mask1 = singa.GTFloat(m,0)
+        mask11=singa.__mul__(mask1,b)
+        mask = singa.__add__(mask00,mask11)
+        
+        if training:
+            self.mask = mask
```
nudles: mask is not used anymore?

python/singa/autograd.py
pinpom: yes. I follow https://github.com/tensorflow/tensorflow/blob/master/tensorflow/cc/gradients/math_grad.cc#L782, so the dx_0=mask0; dx_1=mask1

python/singa/autograd.py
pinpom: yes. only used in forward function

nudles: then there is no need to store it in self.mask.

pinpom: yes, I just un-stored 'mask'

PR #511, opened 2019-08-14
SINGA-474 squeeze operator

ShichengChen:


python/singa/autograd.py
```diff
@@ -1583,6 +1583,33 @@ def backward(self, dy):
...
+class Squeeze(Operation):
+    def __init__(self,axis=-1):
+        super(Squeeze, self).__init__()
+        if(type(axis) is int):
+            self.axis=[axis]
+        else:
+            self.axis=axis
+
+
+    def forward(self, x):
```
nudles: what does axis=-1 mean?
pls follow the documentation here https://github.com/onnx/onnx/blob/master/docs/Operators.md#squeeze

ShichengChen: I followed numpy before. axis=-1 means squeeze all dimensions which are equal to one.
I just changed squeeze to onnx version.

PR #514, opened 2019-08-14
tc comprehension integration Ref. SINGA-482

dcslin:
Hi this is not completely ready for merge, pending to fix a testing error.


dcslin:
Hi, this is the working version of tc integration for now, with following notes:
1. tested in the environment specified in the dockerfile `cuda9`
2. it implemented `MatMul`, `Relu` in TC for demonstration and usage is shown in following tests:
` ./bin/test_singa --gtest_filter=*TC*`, and tests are passing.
3. for now only working in Cuda environment

P.S. `softmax` is implemented but still checking why the output is not correct.

dcslin:
TODO:
1. move tc execution code into device `DoExec`, and support CPU
2. package dependencies as conda package
3. integer support

dcslin:
Integration is done except support CPU.
CPU support development is done, however the behavior was not expected even in clean TC env.
CPU backed TC return zero tensor as shown in https://gist.githubusercontent.com/dcslin/f3ed411012c144163f3bedbe621257f6/raw/9f6af3f5e62b5bb52a58c087967c813b3acd8872/tensor-comprehension-cpu-errors

dcslin:
Ready for review.

PR #515, opened 2019-08-14
SINGA-474 Add gemm operator

nudles:


python/singa/autograd.py
```diff
@@ -2316,3 +2315,37 @@ def backward(self, dy):
...
+            B = singa.DefaultTranspose(B)
+        if training:
+            self.A = A
+            self.B = B
+        
+        singa.MultWithScale(self.alpha, A, B, self.beta, C)
+        return C
+
+    def backward(self, dY):
+        dC = singa.MultFloat(dY, self.beta)
```
chrishkchris: Is the backward the same when transA and transB =1 or 0?

PR #517, opened 2019-08-15
SINGA-474 squeeze operator

ShichengChen:


python/singa/autograd.py
```diff
@@ -2194,6 +2194,30 @@ def log(x):
...
 
 
+class Squeeze(Operation):
+    def __init__(self,axis=[]):
+        super(Squeeze, self).__init__()
+        self.axis=axis
+
+    def forward(self, x):
+        self.cache=x.shape()
+        cur = list(self.cache)
```
nudles:
```
newshape = []
for i in self.axis:
   assert self.cache[i] == 1, "the length of axis {} is {}, which should be 1".format(i, self.cache[i])
for d in self.cache:
   if d > 1:    
      newshape.append(d)
```

ShichengChen: sorry, I fixed the bug followed your code.

PR #520, opened 2019-08-15
SINGA-474 add operator

ShichengChen:
make add operator support broadcast forward and backward.
still keep add_bias, to make old code can be run

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -612,8 +612,10 @@ class Add(Operation):
         super(Add, self).__init__()
 
     def forward(self, a, b):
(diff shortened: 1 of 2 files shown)


python/singa/autograd.py
```diff
@@ -611,10 +612,17 @@ def __init__(self):
         super(Add, self).__init__()
 
     def forward(self, a, b):
+        self.shape0=list(a.shape())
+        self.shape1=list(b.shape())
```
nudles: can it handle the case that a is a matrix and b is a vector?

PR #522, opened 2019-08-16
SINGA 475 - add And operator

pinpom:


python/singa/autograd.py
```diff
@@ -2319,3 +2319,24 @@ def backward(self, dy):
 def max(a,b):
     return Max()(a,b)[0]
 
+
+class And(Operation):
+    def __init__(self):
+        super(And, self).__init__()
+
+    def forward(self, a, b):
+        m = singa.__div__(a,b)
```
nudles: there will be errors if b has 0s.

pinpom: i assume a/0 = un-defined. so in both cases, mask0 = singa.GEFloat(m,1) & mask1 = singa.LEFloat(m,1) return 0 for any element of a/0. the test also passed when b has 0s (b = np.array([0, -0.3, 0, 0.1, 0.5, 0.9])). 
I am not sure if in computing there is error when trying to compute a/0? please help to advise.

pinpom:
closed due to duplicate #528

PR #524, opened 2019-08-18
SINGA-474 prelu,add,equal,selu,elu operator

ShichengChen:


python/singa/autograd.py
```diff
@@ -608,21 +608,177 @@ def backward(self, dy):
...
     def forward(self, a, b):
+        #up till now, the dimensions of tensor a and b should less than 3
+        self.shape0=list(a.shape())
+        self.shape1=list(b.shape())
+        assert(len(self.shape0) <= 2 and len(self.shape1) <= 2),"up till now, the dimensions of tensor a and b should less than 3"
         return singa.__add__(a, b)
 
     def backward(self, dy):
-        return dy, dy
+        if(type(dy)==float):return dy,dy
```
nudles: this implementation assumes that the a has the same shape as dy?
then need to assert it.

python/singa/autograd.py
```diff
@@ -608,21 +608,177 @@ def backward(self, dy):
...
+            a CTensor for the result
+        """
+        #f(x) = alpha * (exp(x) - 1.) for x < 0, f(x) = x for x >= 0
+        if training:
+            self.input = x
+        x1 = singa.LTFloat(x, 0.0)
+        x1 = singa.__mul__(x, x1)
+        x1 = singa.MultFloat(singa.SubFloat(singa.Exp(x1),1.0),self.alpha)
+        x2 = singa.ReLU(x)
+        x1 = singa.__add__(x1, x2)
```
nudles: can use ```x1+=x2```?

python/singa/autograd.py
```diff
@@ -608,21 +608,177 @@ def backward(self, dy):
...
+            a tuple for dx
+        """
+        dx1mask = singa.LTFloat(self.input, 0.0)
+        dx1 = singa.MultFloat(singa.Exp(self.input), self.alpha)
+        dx1 = singa.__mul__(dx1mask, dx1)
+
+        dx2mask = singa.GEFloat(self.input, 0.0)
+
+        dx = singa.__add__(dx1, dx2mask)
+        return singa.__mul__(dy, dx)
```
nudles: can use dx *= dy?

test/python/test_operation.py
```diff
@@ -1587,25 +1587,148 @@ def test_min_gpu(self):
...
+            if(gpu):
+                dy.to_device(gpu_dev)
+                x.to_device(gpu_dev)
+            dx = result.creator.backward(dy.data)
+            np.testing.assert_array_almost_equal(tensor.to_numpy(result), y, decimal=5)
+            np.testing.assert_array_almost_equal(tensor.to_numpy(tensor.from_raw_tensor(dx)), grad, decimal=5)
+        test_helper(False)
+        test_helper(True)
+
+    @unittest.skipIf(not singa_wrap.USE_CUDA, 'CUDA is not enabled')
```
nudles: will it skip the cpu test as well?

ShichengChen: I just changed it to if-else checks inside unit test functions

python/singa/autograd.py
ShichengChen: yes

python/singa/autograd.py
ShichengChen: yes

python/singa/autograd.py
ShichengChen: yes

PR #528, opened 2019-08-23
SINGA 475 - add logical operator: and, or, xor, not & negative, reciprocal

pinpom:


python/singa/autograd.py
```diff
@@ -2522,3 +2523,144 @@ def backward(self, dy):
 def max(a,b):
     return Max()(a,b)[0]
 
+
+class And(Operation):
+    def __init__(self):
+        super(And, self).__init__()
+
+    def forward(self, a, b):
+        m = singa.__div__(a,b)
```
joddiy: This part cannot pass the cpu test case, plz use this.
```
        m = singa.__mul__(a, b)
        cur = singa.PowFloat(singa.Sign(m), 2)
```

python/singa/autograd.py
```diff
@@ -2522,3 +2523,144 @@ def backward(self, dy):
...
+    return And()(a,b)[0]
+
+
+class Or(Operation):
+    def __init__(self):
+        super(Or, self).__init__()
+
+    def forward(self, a, b):
+        #find equal element-wise
+        m = singa.__sub__(a,b)
```
joddiy: not clear, plz use this:
```
        m = singa.__add__(singa.PowFloat(singa.Sign(a), 2.0), singa.PowFloat(singa.Sign(b), 2.0))
        cur = singa.Sign(m)
```

python/singa/autograd.py
```diff
@@ -2522,3 +2523,144 @@ def backward(self, dy):
...
+    return Not()(x)[0]
+
+
+class Xor(Operation):
+    def __init__(self):
+        super(Xor, self).__init__()
+
+    def forward(self, a, b):
+        #find element with value =0
+        m0 = singa.__mul__(a,b)
```
joddiy: not clear, plz use this:
```
        m = singa.__sub__(singa.PowFloat(singa.Sign(a), 2.0), singa.PowFloat(singa.Sign(b), 2.0))
        cur = singa.PowFloat(singa.Sign(m), 2.0)
```

python/singa/autograd.py
pinpom: i think the equation is not correct if a = -b (element-wise)

joddiy: if a = -b, the Signs of them are, 1, and -1, and then, you take a square on it, the result will be 1 and 1, and you add them together, is 2, take the Sign again, you get a 1 finally.

pinpom: ok got it. just changed the code already.

PR #529, opened 2019-08-24
SINGA-484 Code analysis with LGTM

moazreyad:
Adding code analysis badges to README.

### DIFF
--- a/include/singa/io/transformer.h
+++ b/include/singa/io/transformer.h
@@ -56,12 +56,12 @@ class ImageTransformer: public Transformer {
 
   Tensor Apply(int flag, Tensor& input) override;
 
-  const bool featurewise_center() const { return featurewise_center_; }
-  const bool featurewise_std_norm() const { return featurewise_std_norm_; }
-  const bool horizontal_mirror() const { return horizontal_mirror_; }
-  const int resize_height() const { return resize_height_; }
-  const int resize_width() const { return resize_width_; }
-  const float rescale() const { return rescale_; }
+  bool featurewise_center() const { return featurewise_center_; }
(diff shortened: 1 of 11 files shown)


include/singa/io/transformer.h
```diff
@@ -56,12 +56,12 @@ class ImageTransformer: public Transformer {
 
   Tensor Apply(int flag, Tensor& input) override;
 
-  const bool featurewise_center() const { return featurewise_center_; }
-  const bool featurewise_std_norm() const { return featurewise_std_norm_; }
-  const bool horizontal_mirror() const { return horizontal_mirror_; }
-  const int resize_height() const { return resize_height_; }
-  const int resize_width() const { return resize_width_; }
-  const float rescale() const { return rescale_; }
+  bool featurewise_center() const { return featurewise_center_; }
```
nudles: Does the code analysis suggest to remove the const?
Where can we find all the suggestions?
Thanks.

moazreyad: Yes, LGTM suggested to the remove the const because it is useless. More information can be found [here](https://lgtm.com/rules/2156200647/). 

There are more alerts that I could try to fix, but because the [testing pipeline is not working](https://issues.apache.org/jira/browse/SINGA-488), I can't be sure that my fixes will not cause other problems.

All the alerts can be found in this [link](https://lgtm.com/projects/g/apache/incubator-singa/alerts/?mode=list), which is also available in the Jira ticket [484](https://issues.apache.org/jira/browse/SINGA-484)

In this pull request, I propose to add the code analysis badges to SINGA readme and I fixed few alerts. But there are many more to be fixed.

PR #533, opened 2019-09-10
fixed operation add assertion for convolution

dcslin:
this is required to run the `autograd/resnet.py`

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -645,7 +645,10 @@ class Add(Operation):
         #up till now, the dimensions of tensor a and b should less than 3
         self.shape0=list(a.shape())
         self.shape1=list(b.shape())
-        assert(len(self.shape0) <= 2 and len(self.shape1) <= 2),"up till now, the dimensions of tensor a and b should less than 3"
+
+        # fix for convolution, tensor has 4 dims
+        assert( (len(self.shape0) <= 2 and len(self.shape1) <= 2) or (self.shape0 == self.shape1) ),"up till now, the dimensions of tensor a and b should less than 3"
+        
         return singa.__add__(a, b)
 
     def backward(self, dy):


dcslin:
2nd commit fixed that "cudnn softmax is not used" mentioned in https://github.com/apache/incubator-singa/pull/468#issuecomment-521247206

dcslin:
after the fix, testing on `python/resnet.py` shows similar result:
Before:
```
100%|#######################################################################################################################################################################| 100/100 [00:27<00:00,  3.68it/s]
Throughput = 58.903093484671295 per second
Total=0.27163259267807005, forward=0.028347208499908447, softmax=0.10690386295318603, backward=0.13638152122497554, sgd=0.08405628204345703
```

After:
```
100%|#######################################################################################################################################################################| 100/100 [00:27<00:00,  3.69it/s]
Throughput = 59.02303205471914 per second
Total=0.2710806179046631, forward=0.02830531120300293, softmax=0.10648007154464721, backward=0.13629523515701297, sgd=0.08383089303970337
```

src/core/tensor/tensor_math_cpp.h
```diff
@@ -868,6 +868,25 @@ void RowMax<float, lang::Cpp>(const Tensor& in, Tensor *out, Context *ctx) {
   }
 }
 
+template <>
+void SoftMax<float, lang::Cpp>(const Tensor& in, Tensor* out, Context *ctx) {
```
nudles: do we need to move SoftMax into tensor_math_cpp.h?
If not, please delete the changes.

dcslin: Hi @nudles , it is not implemented in cpp yet. So we need to move this from `tensor.cc` into `tensor_math_cpp.h`. If not, when calling softmax with cpp, it will through `not impl` fatal error.

dcslin:
Hi @nudles , this is ready for merge.
thanks to @chrishkchris

PR #538, opened 2019-09-26
[WIP] add Cudnn mode as optional param to batch norm and softmax

dcslin:


dcslin:
split to https://github.com/apache/incubator-singa/pull/540 and https://github.com/apache/incubator-singa/pull/541

PR #541, opened 2019-10-03
SINGA-476 added 4d test on batchnorm

dcslin:
added batchnorm test adapted from `/test/python/test_onnx_backend.py` and test passed.
Hi @joddiy , If current Batchnorm API could fulfil ONNX batchnorm requirement, shall we keep the current API unchanged? Kindly advise. Thanks. 
Ref. SINGA-476


joddiy:
the return of _CpuBatchNormForwardInference_ should be y, var, mean, not just y.
could you fix it?

dcslin:
HI @joddiy , could you please help to review this test on batchnorm value? which also showing that the required params/output by ONNX are given.

joddiy:
hi, shicong, the result still has a little error, please check with the following case:
```
        def _batchnorm_test_mode(x, s, bias, mean, var, epsilon=1e-5):  # type: ignore
            dims_x = len(x.shape)
            dim_ones = (1,) * (dims_x - 2)
            s = s.reshape(-1, *dim_ones)
            bias = bias.reshape(-1, *dim_ones)
            mean = mean.reshape(-1, *dim_ones)
            var = var.reshape(-1, *dim_ones)
            return s * (x - mean) / np.sqrt(var + epsilon) + bias

        # input size: (1, 2, 1, 3)
        x = np.array([[[[-1, 0, 1]], [[2, 3, 4]]]]).astype(np.float32)
        s = np.array([1.0, 1.5]).astype(np.float32)
        bias = np.array([0, 1]).astype(np.float32)
        mean = np.array([0, 3]).astype(np.float32)
        var = np.array([1, 1.5]).astype(np.float32)
        y = _batchnorm_test_mode(x, s, bias, mean, var).astype(np.float32)
```

dcslin:
HI @joddiy , could you share how did you call singa api? the test you mentioned is tested in here https://github.com/apache/incubator-singa/pull/541/files#diff-5e6248de6199ea6f777c73d3ea2643a5R62

for the `test_mode`, kindly use this api `GpuBatchNormForwardInference`

joddiy:
ready to merge

dcslin:
Hi @nudles , @joddiy , conflicts are resolved.

PR #554, opened 2019-10-27
[bug]batchnorm running mean var not updated

dcslin:
Hi team,
Shouldn't `autograd` `batchnorm`  class update its running_mean and running_var from c++ backend? I could not find anywhere `running_mean/var` are updated.

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -1455,7 +1455,7 @@ class _BatchNorm2d(Operation):
 
                 self.cache = (x, scale, mean, var, y, bias)
             else:
-                y, mean, var = singa.GpuBatchNormForwardTraining(
+                y, mean, var, self.running_mean, self.running_var = singa.GpuBatchNormForwardTraining(
                     self.handle, x, scale, bias, self.running_mean, self.running_var
                 )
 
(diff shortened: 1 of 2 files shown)


chrishkchris:
I think this is pass by reference (the running mean and variance singa tensors' blocks are fed to the function cudnnBatchNormalizationForwardTraining, so there is no need to return the running mean and variance singa tensors (because the _BatchNorm2d already have them)?

chrishkchris:
Oh I see. The running mean and var are updated using NOT in place assignment.
running_mean = running_mean * bnh.factor + mean * (1 - bnh.factor);
If we change it to in place operation, there is no need to return, i.e.
running_mean*=bnh.factor;
Axpy(1-bnh.factor,mean,running_mean)

dcslin:
Thank you @chrishkchris 

I tested and currently in python the running mean is not updated after `GpuBatchNormForwardTraining` returns.
Another thing to note is that, SWIG has this `apply OUTPUT` which might solve this. http://www.swig.org/Doc1.3/Arguments.html
I looking into this.

chrishkchris:
Yes, I see why the running mean is not updated. This is because
running_mean = running_mean * bnh.factor + mean * (1 - bnh.factor);
It generates a new tensor to be stored in the variable running mean, since it is NOT in place operation. this is why the  original running mean is not updated
So if we use in place operation the Original running mean tensor can be updated.

dcslin:
Hi @chrishkchris thank you for the help. This is false alarm. the code is working as @chrishkchris described.

PR #568, opened 2019-12-16
SINGA-487 Update the descriptions to DistOpt API

chrishkchris:


chrishkchris:
updated based on google python style guide http://google.github.io/styleguide/pyguide.html

PR #570, opened 2019-12-23
Tensor naming convention review

dcslin:
1. added additional word to indicate mutation on underline data
2. add inplace flag to indicate mutation on underline data
3. updated cpp `sum()` on axis
4. no change to python `add_column()` or similar as it could be done by `broadcast` and could be deprecated.

### DIFF
--- a/python/singa/tensor.py
+++ b/python/singa/tensor.py
@@ -184,6 +184,9 @@ class Tensor(object):
         return t
 
     def reset_like(self, t):
+        self.reset_like_(t)
+
+    def reset_like_(self, t):
         '''Reset the shape, dtype and device as the given tensor.
 
         Args:
@@ -233,6 +236,9 @@ class Tensor(object):
         return self.data.L1()
 
     def set_value(self, x):
+        self.set_value_(x)
+
+    def set_value_(self, x):
         '''Set all elements of the tensor to be the give value.
 
         Args:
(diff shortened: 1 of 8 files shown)


python/singa/tensor.py
```diff
@@ -232,14 +232,18 @@ def l1(self):
...
         Args:
             x (float), a float value to be set to all elements.
         '''
         # assert type(x) == float, 'set value only accepts float input'
         # if isinstance(x, float):
+        if not inplace:
+            # return new tensor filled with value
+            raise NotImplementedError
+
         self.data.SetFloatValue(float(x))
```
nudles: return self if inplace is false; otherwise, return the new tensor?

python/singa/tensor.py
```diff
@@ -352,74 +356,86 @@ def deepcopy(self):
...
             low (float): the lower bound
             high (float): the hight bound
         '''
+        if not inplace:
+            # return new tensor
+            raise NotImplementedError
+
         singa.Uniform(float(low), float(high), self.data)
 
     def add_column(self, v):
```
nudles: use decorator  ```@deprecated ```

nudles: print the the new function name in the msg.

dcslin: thanks for the review, this is fixed.

python/singa/tensor.py
dcslin: thanks for the review, this is fixed

dcslin:
Hi @nudles, thanks for the review, this is ready for merge.

python/singa/tensor.py
dcslin: reference https://github.com/apache/singa/pull/570/commits/203fcb91cd590d3436ce31a68844d8f47817be9c

python/singa/tensor.py
dcslin: reference https://github.com/apache/singa/pull/570/commits/8cc44683abfb21be284a751f501ba31eb4737022

PR #580, opened 2020-01-22
[WIP] SINGA-505 An experiment for buffering operations

chrishkchris:
Here I show the progress for my branch which is an experiment for buffering operations.

In the experiment, now I can buffer forward loop operations until the end of SoftMax_cross_entropy, I still need to work on backward operations.


chrishkchris:
```
ubuntu@ip-172-31-21-156:~/singa/examples/autograd$ python3 mlp3.py
train_data_shape: (400, 2)
train_label_shape: (400, 2)
EnterEXEC
EnterEXEC
EnterBuffStart, 2
EnterBuffExit
EnterBuffStart, 1
EnterBuffExit
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterBuffStart, 4
EnterBuffExit
EnterBuffStart, 3
EnterBuffExit
EnterBuffStart, 2
EnterBuffExit
EnterBuffStart, 1
EnterBuffExit
EnterEXEC
EnterEXEC
EnterBuffStart, 2
EnterBuffExit
EnterBuffStart, 1
EnterBuffExit
[10. 10.]
EnterEXEC
enter Addrow
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
enter Addrow
EnterEXEC
EnterEXEC
EnterBuffStart, 7
EnterBuffExit
EnterBuffStart, 6
EnterBuffExit
EnterBuffStart, 5
EnterBuffExit
EnterBuffStart, 4
EnterBuffExit
EnterBuffStart, 3
EnterBuffExit
EnterBuffStart, 2
EnterBuffExit
EnterBuffStart, 1
EnterBuffExit
EnterEXEC
EnterEXEC
EnterBuffStart, 2
EnterBuffExit
EnterBuffStart, 1
EnterBuffExit
[30. 30.]
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterBuffStart, 5
enter cpp softmax
immediately ops
immediately ops
enter Addrcolumn
immediately ops
immediately ops
immediately ops
immediately ops
immediately ops
immediately ops
immediately ops
EnterBuffExit
EnterBuffStart, 4
enter cpu crossentropyEnterBuffExit
EnterBuffStart, 3
EnterBuffExit
EnterBuffStart, 2
EnterBuffExit
EnterBuffStart, 1
EnterBuffExit
EnterEXEC
EnterEXEC
EnterBuffStart, 2
EnterBuffExit
EnterBuffStart, 1
EnterBuffExit
0.6931472
start backward loop
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterEXEC
EnterBuffStart, 16
EnterBuffExit
EnterBuffStart, 15
Segmentation fault (core dumped)
```
So now I am debugging the backward loop...

dcslin:
Hi, this is regarding the `SumAll` in this PR. Referring to numpy, when no axis is given, `numpy.sum` returns a scalar value. Referring to torch, `torch.sum()` always return tensor. This difference is because there is no tensor in numpy.

I suppose the ideal way for singa would be deprecating `float Sum()` and keeping `Tensor Sum()`.

However this breaks the current code.

The workaround might be extending `Tensor Sum(Tensor in, int axis)` to support arbitrary axes like `Tensor Sum(Tensor in, int[] axis)`. Then when we want `Tensor SumAll()`, we could call `Tensor Sum(in, in.shape())`, which effectively sum all (over all axes) and return a Tensor, while not breaking current code.

nudles:
1. It should be better to return a Tensor instance. To keep it compatible, we can add a new function, sum_at() (returns a tensor), and mark the original sum() as deprecated. When we release v3.0, we mark sum_at as deprecated and change the sum() to return a tensor. This is to follow the convention of the X.Y.Z version style.
2. Is it more common to pass axis as an integer than an array? Do we need to do summation along multiple dimensions?

dcslin:
> 1. It should be better to return a Tensor instance. To keep it compatible, we can add a new function, sum_at() (returns a tensor), and mark the original sum() as deprecated. When we release v3.0, we mark sum_at as deprecated and change the sum() to return a tensor. This is to follow the convention of the X.Y.Z version style.
> 2. Is it more common to pass axis as an integer than an array? Do we need to do summation along multiple dimensions?

Hi @nudles , regarding 1:
currently in Dev:
```
Tensor SumAll(const Tensor &in);
Tensor Sum(const Tensor &in, const int axis);
SType Sum(const Tensor &in);
```
do you mean we should change to:
```
V2.x.x
Tensor SumAll(const Tensor &in);
Deprecated{Tensor Sum(const Tensor &in, const int axis);}
SType Sum(const Tensor &in);
Tesnor SumAt(const Tensor &in, const int axis);
```
and at V3.0
```
V3.0
Tensor SumAll(const Tensor &in);
Tensor Sum(const Tensor &in, const int axis);
Tensor Sum(const Tensor &in);
Deprecated{Tesnor SumAt(const Tensor &in, const int axis);}
```

As mentioned by Chris, SumAll is already merged, and everything should work now... My comment was not aware of that.

Maybe we could just change `SType Sum()` to `SType SumValue()`.

Regarding 2. passing axis as array is just an API supported by both numpy, torch.

nudles:
We can have the following c++ functions in V2.x
```
/// deprecated
Tensor SumAll(const Tensor &in);
/// deprecated
SType Sum(const Tensor &in);
Tensor Sum(const Tensor &in, const int axis=-1);
Tensor Sum(const Tensor &in, const int[] axis);
```
For the Python side
```
def sum(in, axis=-1)  # the axis could be an integer or an array
@deprecated 
def SumAsFloat(in)
```

In V3.0, we remove the deprecated functions.

dcslin:
Hi @nudles , I was implementing the these fn and found out it is already done in python level. 
https://github.com/apache/singa/blob/master/python/singa/tensor.py#L845

Hi @joddiy , could you please advise if Sum( with axis) at python api cater your need?

chrishkchris:
Since I have asked @XJDKC to open a new PR #626 to replace this PR (this PR is for experimental only), now I closed this PR and replaced by #626

PR #587, opened 2020-01-30
SINGA-504 Add Gemm operator for autograd and onnx

joddiy:


python/singa/autograd.py
```diff
@@ -2760,3 +2760,107 @@ def backward(self, dy):
...
+        Args:C: 
+            tensor(optional), Optional input tensor C. If not specified, the computation is done as if C is a scalar 0. The shape of C should be unidirectional broadcastable to (M, N).
+        Returns: 
+            tensor, the output
+        """
+        _A = singa.DefaultTranspose(A) if self.transA == 1 else A
+        _B = singa.DefaultTranspose(B) if self.transB == 1 else B
+        if training:
+            self.inputs = (_A, _B, C)
+        tmpM = singa.MultFloat(singa.Mult(_A, _B), self.alpha)
```
nudles: Did you try [MultiWithScale](https://github.com/apache/singa/blob/master/src/api/core_tensor.i#L319)?
Or the implementation in this [PR](https://github.com/apache/singa/pull/515/files#diff-a236c495700ad62b39e28a4ece74711fR2318)

test/python/test_operation.py
```diff
@@ -2745,5 +2745,58 @@ def test_prelu_broadcast_cpu(self):
             np.testing.assert_array_almost_equal(tensor.to_numpy(tensor.from_raw_tensor(dx0)), grad0, decimal=5)
             np.testing.assert_array_almost_equal(tensor.to_numpy(tensor.from_raw_tensor(dx1)), grad1, decimal=5)
 
+    def test_gemm(self):
+        devs = [cpu_dev, gpu_dev]
```
nudles: what if the machine does not have gpu?

chrishkchris: > what if the machine does not have gpu?

Yes, that's true.
Last time I run the test_operation.py with USE_CUDA=0 in the master branch, the test_operation.py cannot run before removing the gpu_dev related things.

python/singa/autograd.py
joddiy: The MultiWithScale needs 5 params(alpha, A, B, beta, C), if the user doesn't input the C (C is None), I guess I still need to init a C with zero values...

joddiy: MultiWithScale needs 5 inputs, (alpha, A, B, beta, C), if the user doesn't input a C, which means C is none, it still needs to init a C with all zero values. maybe we should optimize the MultiWithScale?

nudles: how to optimize it?
I think MultiWithScale is more efficient as it calls the cpp/cuda gemm directly instead of doing the matrix multiplication and addition separately..

joddiy: I just try this MultiWithScale, but it cannot support the Unidirectional Broadcasting for bias. https://github.com/onnx/onnx/blob/master/docs/Broadcasting.md

joddiy: and also, please check the test cases for gemm, https://github.com/onnx/onnx/blob/master/onnx/backend/test/case/node/gemm.py#L27-L108

nudles: I see. This PR can pass the tests?
Pls fix the conflicts and format the code.

joddiy: Yes, since shicong has re-implement the add and mul of singa to support it.
got it.

joddiy:
@nudles has formatted the code and is ready to merge.

PR #589, opened 2020-02-01
hotfix on dev branch (remove of duplicate variable)

chrishkchris:
Maybe there are some errors in DNNL needed to be fixed before functioning correctly.

I currently have solved a build error due to reassign of variable, so now I can built the code. The c test case test_singa.o and python mlp.py can now pass.

There are two more errors I encountered but I did not solve yet
1. mnist_cnn (after changing to cpu), it hangs.
2. test_operation.py
```
ubuntu@ip-172-31-24-48:~/singa/test/python$ python3 test_operation.py
.......................................Segmentation fault (core dumped)
```

PR #590, opened 2020-02-03
fix for both USE DNNL and USE CUDNN are enabled, code format, test ap…

dcslin:
…i all ok


dcslin:
fix for https://github.com/apache/singa/pull/589

dcslin:
FYI, when compiling both USE DNNL and USE CUDNN, `operation handle` baes class create DNNL handle and free DNNL handle even on gpu operations.

PR #592, opened 2020-02-04
fix bugs for some cases of onnx

joddiy:
* We can support the input of clip like [x, min, max], [x, min], [x], cannot support like [x, max]
* Add shape check of auto_pad for conv2d of onnx

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -492,7 +492,7 @@ class Clip(Operation):
         return singa.__mul__(dy, self.mask)
 
 
-def clip(x, min, max):
+def clip(x, min=None, max=None):
     return Clip(min, max)(x)[0]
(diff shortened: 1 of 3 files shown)


nudles:
have you updated the clip function to support [x, min, max], [x, min], [x]?
ready for merge?

joddiy:
> have you updated the clip function to support [x, min, max], [x, min], [x]?
> ready for merge?

Yes, I've already supported the [x, min, max], [x, min], [x]. 

But if we want to support [x, max], because it doesn't give the parameters with the default order, we must change lots of code...

Can merge now.

nudles:
does onnx require to support [x, max]?
I will merge this PR, pls create an issue for [x, max] if necessary.

PR #595, opened 2020-02-10
Clang format for distributed module

chrishkchris:
I suggest format the code in distributed module first, so that the changes in future PR can be tracked more easily.


chrishkchris:
updated on 2/11 using the google style setting in clang-formatter
"clang-format.language.cpp.style": "google"

PR #597, opened 2020-02-11
added tensor index selector

dcslin:


dcslin:
Hi @joddiy , regarding Conv padding, i understand that the missing part is the index selection, e.g.`tensor1[:,:,1:,1:]`, which is added in this PR.

Let me know if further info is required.

@nudles , FYI

joddiy:
thanks, shicong, I'll try it ASAP.

joddiy:
Hi, @dcslin, for the backward of CNN, I guess I should firstly pad the dy as its shape same with the shape before index selection. So, I need a function that can add one column and row at the end or start of the matrix. For example, for "SAME_UPPER" mode, and output shape of [3,3,32,32], I need to pad one column and one row at the end of the dy to make the shape as [3,3,32,32].

dcslin:
> Hi, @dcslin, for the backward of CNN, I guess I should firstly pad the dy as its shape same with the shape before index selection. So, I need a function that can add one column and row at the end or start of the matrix. For example, for "SAME_UPPER" mode, and output shape of [3,3,32,32], I need to pad one column and one row at the end of the dy to make the shape as [3,3,32,32].

hi could it be done in reverse way? add extra padding and make it symmetrical, and trim off one column by using the selection[].

joddiy:
> > Hi, @dcslin, for the backward of CNN, I guess I should firstly pad the dy as its shape same with the shape before index selection. So, I need a function that can add one column and row at the end or start of the matrix. For example, for "SAME_UPPER" mode, and output shape of [3,3,32,32], I need to pad one column and one row at the end of the dy to make the shape as [3,3,32,32].
> 
> hi could it be done in reverse way? add extra padding and make it symmetrical, and trim off one column by using the selection[].

sure, and how to add extra padding? can use singa.add_column and add_row?

dcslin:
this is ready for merge, which is required in ONNX

PR #598, opened 2020-02-12
fix softmax nan problem

XJDKC:
Here I change the algorithm of the cudnnSoftmaxForward function.

The original algorithm was CUDNN_SOFTMAX_FAST which may lead to overflow when the input numbers are too large. Change it to CUDNN_SOFTMAX_ACCURATE will solve this problem.

For example: If cudnn softmax is used in mlp.py, it will lead to this problem.

### DIFF
--- a/src/core/tensor/tensor_math_cuda.h
+++ b/src/core/tensor/tensor_math_cuda.h
@@ -938,7 +938,7 @@ void SoftMax<float, lang::Cuda>(const Tensor& in, Tensor* out, Context* ctx) {
 template <>
 void SoftMaxBackward<float, lang::Cuda>(const Tensor& in, Tensor* out,
                                         const Tensor& fdout, Context* ctx) {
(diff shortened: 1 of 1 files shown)


chrishkchris:
I strongly support this change for two reasons:
1. The current setting fails some network structure/dataset, which make the functions not universal for all applications.
2. SoftMax forward spent only very short time within the CNN training, e.g. resnet50 has many conv layers but only one SoftMax layer, so the time spent in SoftMax is not the bottleneck.

chrishkchris:
sorry, could you also change CUDNN_SOFTMAX_FAST to CUDNN_SOFTMAX_ACCURATE in SoftMaxbackward in tensor_math_cuda.cc ? I am not sure if both needed to be matched

chrishkchris:
thanks. I think it is ready for merge

XJDKC:
Thank you for your reminding.

PR #599, opened 2020-02-17
Enabled linting check in CI

dcslin:


nudles:
I think it will fail to pass the linting test because many files are not formated.

dcslin:
currently the linting.sh check is checking below files:
```
find python/ \
    examples/autograd \
    test/python/ -iname *.py | xargs pylint

find src/api/ \
    src/core/ \
    src/proto/ \
    src/utils/ \
    include/singa/core/ \
    include/singa/utils/ \
    src/model/operation/ \
    include/singa/io/communicator.h \
    src/io/communicator.cc \
    test/singa/ -iname *.h -o -iname *.cc | xargs cpplint --quiet
```
however the problem is there are still a lot of warnings. One way is to customize the linting config. Other way we define the linting check is for reference and not compulsary.

nudles:
Do the warnings fail the travis testing?

dcslin:
> Do the warnings fail the travis testing?

In current setting, travis is simply checking the returned value of our script `linting.sh`.

The main problem here is there are a lot of constraint with standard linting configuration. I doubt that any large scale project could fullfil all the linting constraint.

chrishkchris:
> Do the warnings fail the travis testing?
> 
> In current setting, travis is simply checking the returned value of our script linting.sh.
> The main problem here is there are a lot of constraint with standard linting configuration. I doubt that any large scale project could fullfil all the linting constraint.

@dcslin I suggest you can paste an example of the linting result here, so everyone can understand how many errors are in the linting result that are too hard to resolve

dcslin:
> > Do the warnings fail the travis testing?
> > In current setting, travis is simply checking the returned value of our script linting.sh.
> > The main problem here is there are a lot of constraint with standard linting configuration. I doubt that any large scale project could fullfil all the linting constraint.
> 
> @dcslin I suggest you can paste an example of the linting result here, so everyone can understand how many errors are in the linting result that are too hard to resolve

Hi @chrishkchris , I'd love to, but since the purpose of integrating linting into CI, thus we could just leverage CI and see the output in the travis log by clicking travis build log.

chrishkchris:
> Do the warnings fail the travis testing?
> In current setting, travis is simply checking the returned value of our script linting.sh.
> The main problem here is there are a lot of constraint with standard linting configuration. I doubt that any large scale project could fullfil all the linting constraint.
> 
> @dcslin I suggest you can paste an example of the linting result here, so everyone can understand how many errors are in the linting result that are too hard to resolve
> 
> Hi @chrishkchris , I'd love to, but since the purpose of integrating linting into CI, thus we could just leverage CI and see the output in the travis log by clicking travis build log.

Yes, I saw you were setting both linting to show only critical one, and I think that would work well so I removed my comment

chrishkchris:
@dcslin 
I saw one interesting error:
https://travis-ci.org/apache/singa/jobs/653724760#L2049
The error is caused because there is no long data type in python 3.X, long data type was used in python 2.X.
https://stackoverflow.com/questions/2104884/how-does-python-manage-int-and-long
However, this error will not cause any runtime error in python 3.X because the whole line is 
`if isinstance(repeats, int) or isinstance(repeats, long):`

nudles:
> 
> 
> > Do the warnings fail the travis testing?
> 
> In current setting, travis is simply checking the returned value of our script `linting.sh`.
> 
> The main problem here is there are a lot of constraint with standard linting configuration. I doubt that any large scale project could fullfil all the linting constraint. So the second problem is to find a balance between linting on everything and linting on nothing.
> 
> Thus I set both linting to show only critical one for now

Then Travis will always pass as long as the linting script can run without errors no matter whether the linting is passed or not?

dcslin:
> > > Do the warnings fail the travis testing?
> > 
> > 
> > In current setting, travis is simply checking the returned value of our script `linting.sh`.
> > The main problem here is there are a lot of constraint with standard linting configuration. I doubt that any large scale project could fullfil all the linting constraint. So the second problem is to find a balance between linting on everything and linting on nothing.
> > Thus I set both linting to show only critical one for now
> 
> Then Travis will always pass as long as the linting script can run without errors no matter whether the linting is passed or not?

`linting.sh` return the returned value from `pylint`/`cpplint`
Travis checks `linting.sh` returned value which is effectively the returned value of  `pylint`/`cpplint`
So if `pylint`/`cpplint` return non 0, Travis fails
if `pylint`/`cpplint` return 0, Travis pass

moazreyad:
I am not sure if linting should be in the build.sh file. This file is executed to test the build of the code, while linting is not related to the build test. It is part of "static code analysis". I would propose to have a new test machine in ".travis.yml" for the static code checks such as linting.

Something like:

  - os: linux
    dist: bionic
    compiler: gcc
    name: "Ubuntu 18.04 (Static Analysis)" 
    script:
        - bash -ex tool/linting/linting.sh

and this machine will run only the script of linting (without building).

Otherwise, this pull request will run the linting four times with each build which is not necessary. 

what do you think?

nudles:
> 
> 
> I am not sure if linting should be in the build.sh file. This file is executed to test the build of the code, while linting is not related to the build test. It is part of "static code analysis". I would propose to have a new test machine in ".travis.yml" for the static code checks such as linting.
> 
> Something like:
> 
>     * os: linux
>       dist: bionic
>       compiler: gcc
>       name: "Ubuntu 18.04 (Static Analysis)"
>       script:
>       - bash -ex tool/linting/linting.sh
> 
> 
> and this machine will run only the script of linting (without building).
> 
> Otherwise, this pull request will run the linting four times with each build which is not necessary.
> 
> what do you think?

+ 1. Agree.

dcslin:
tesing...

dcslin:
@moazreyad Thank you for the advice. I made linting a new job.

dcslin:
I have updated linting config. Currently, when linting fails, travis fails.

Should we fix the linting errors before merging this commit? otherwise, following PRs will be blocked.

Cpp linting errors:
https://travis-ci.org/apache/singa/jobs/654752766?utm_medium=notification&utm_source=github_status

Python linting errors:
https://travis-ci.org/apache/singa/jobs/654752765?utm_medium=notification&utm_source=github_status

nudles:
yes. pls fix the linting errors in this PR.

dcslin:
ready for merge @nudles  thanks

moazreyad:
but why you used two machines for the static analysis? May be one is enough with two scripts for cpp and python:

```
  - os: linux
    stage: linting
    dist: bionic
    compiler: gcc
    name: "Ubuntu 18.04 (Static Analysis)"
    script:
      - bash -ex tool/linting/cpp.sh
      - bash -ex tool/linting/py.sh
```

The static analysis machine can be used also for RAT test, similar to [Apache HAWQ](https://github.com/apache/hawq/blob/master/.travis.yml#L83) and [Apache servicecomb](https://github.com/apache/servicecomb-java-chassis/blob/master/scripts/travis.sh#L34), may be something like:
```

  script:
      - bash -ex tool/linting/cpp.sh
      - bash -ex tool/linting/py.sh
      - bash -ex tool/rat/rat.sh

```
and in the rat.sh, something like:

```
mvn apache-rat:check
if [ $? != 0 ]; then
            echo "${red}Rat check failed.${reset}"
            exit 1
fi
```
This automatic check of licence headers with RAT will make the release process faster. Because now we have to do it manually before the release.

PR #601, opened 2020-02-18
SINGA-504 Add Gemm operator for autograd and onnx

joddiy:
Since I delete the last PR by accident, so I re-commit a new one.
**By the way, the format tool has a bug, even when I select some lines of code and chose to "format selection", the tool even will format all lines, and cause lots of conflicts.**
-> **Need to set the "editor.formatOnSave" as false**

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -2762,6 +2762,7 @@ def reciprocal(x):
     return Reciprocal()(x)[0]
 
 
+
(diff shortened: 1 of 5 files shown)


nudles:
is this PR to be merged into master?

joddiy:
> is this PR to be merged into master?

yes, it's same with https://github.com/apache/singa/pull/600

just re-format the code

PR #602, opened 2020-02-18
CNN add pad mod

joddiy:
the cnn's forward can support SAME_UPPER and SAME_LOWER now


nudles:
Is this still a draft?

joddiy:
> Is this still a draft?

still working on it, trying to fix the backward, now we need to add a column zeros at the end of the dy.
I am discussing with shicong.

PR #607, opened 2020-02-21
Add concat py tensor api

dcslin:


dcslin:
this is ready for merge @nudles

dcslin:
This is fixed. Much thanks

PR #610, opened 2020-02-27
Fix a bug: change #IF USE_CUDNN to #IF USE_DNNL

chrishkchris:
@XJDKC Thanks RULIN for finding out the bug during the cmake operation


chrishkchris:
Without this fix, it cannot build if USE_DNNL=FALSE

PR #612, opened 2020-02-28
Singa 507 _Conv2d and _Pooling2d support odd padding

joddiy:
Since onnx requests same padding mode, we need to optimize conv and pooling these 2 operators.

FYI:

> auto_pad must be either NOTSET, SAME_UPPER, SAME_LOWER or VALID. Where default value is NOTSET, which means explicit padding is used. SAME_UPPER or SAME_LOWER mean pad the input so that the output spatial size match the input.In case of odd number add the extra padding at the end for SAME_UPPER and at the beginning for SAME_LOWER. VALID mean no padding.

This JR will finish:

- lower-level ops(_Conv2d and _Pooling2d) support odd padding mode, for example, if padding is [2,5,2,5], it will crop to a normal padding with [2,2,2,2], and a odd padding with [0,3,0,3]. The odd padding will be added to x tensor firstly and then do normal process with normal padding, then in the backward, the odd padding will be cropped.
- upper-level ops support four bins padding(i.e., (1,2,1,2,)) and also SAME_UPPER, SAME_LOWER for both 2d and 1d input.
- For SAME_UPPER, SAME_LOWER, can do shape checking, and if not set shape, can compute the correct shape.
- Test cases.

### DIFF
--- a/python/singa/autograd.py
+++ b/python/singa/autograd.py
@@ -24,6 +24,7 @@ import numpy as np
 import math
 
 from singa import tensor
+from singa import utils
 from .tensor import Tensor
 from . import singa_wrap as singa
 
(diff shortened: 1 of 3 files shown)


joddiy:
this PR depends on #607 #597

PR #619, opened 2020-03-03
added ceil to tensor, and tests

dcslin:


dcslin:
Hi @joddiy , ceil is added, please refer to https://github.com/apache/singa/blob/5b784f8d90d236420c2d705853fe80b8761f9537/test/python/test_tensor.py#L327 for usage

PR #620, opened 2020-03-03
SINGA-508 Add Apache RAT licence check with Github Actions

moazreyad:
Currently it fails because:

1. It dose not use [rat-excludes](https://github.com/apache/singa/blob/master/rat-excludes). Also this file is not complete. There are many files and directories that must be excluded because they are mentioned in [LICENSE](https://github.com/apache/singa/blob/master/LICENSE). We may use the [excludes option](https://creadur.apache.org/rat/apache-rat-plugin/rat-mojo.html#excludes) to fix this.

2. It executes after `git submodule update ` which downloads Rafiki code. The code has no licence header. We can add the licence to Rafiki, add Rafiki to the excluded directories from check, or run the check before downloading Rafiki code.


moazreyad:
The RAT test pass now after fixing with "exculdes" option. 

This PR can be merged, although I think it is better to combine the RAT test with the linting tests of #599 in one "Static Analysis" test to reduce TravisCI check time.

nudles:
Are you going to move combine it with the linting test?

moazreyad:
I am not sure how to combine the two pull requests together. May be we can merge one of them first, then we update and merge the second later.

For example, we may merge this PR with apache master, then update the apache dev branch from the master and merge the linting PR. or do you propose a better way?

nudles:
the linting PR is already merged into dev..
can you cancel this PR, update the linting script in dev and submit the PR to dev?

moazreyad:
How can I update the dev? should I create a temporary branch at the apache repository for this patch?

I tried to make this PR from moazreyad:SINGA-508 to apache:dev but there are conflicts.

moazreyad:
> Are you going to move combine it with the linting test?

Let's not combine RAT with lint now because it may take some time to make the correct PRs and this issue is small and not urgent now.

.travis.yml
```diff
@@ -37,6 +37,12 @@ matrix:
     dist: bionic
     compiler: gcc
     name: "Ubuntu 18.04 (C++)" 
+  - os: linux
+    dist: bionic
```
nudles: I think we can remove the travis check since  .github/workflows/rat.yml already does the check.

moazreyad: Yes, travis check is removed. This PR can be a first step in moving from travis to github actions. Then later we can move linting and other scripts also.
Can we merge it directly to master? or we still need to merge it with dev first?

nudles:
I think it is better to move it into dev first.
we will merge dev into master before we release the next version.

moazreyad:
Ok. I changed the branch to dev. I hope it is fine now and can be merged.

In this PR, I only changed:
.github/workflows/rat.yml 
java/pom.xml 
tool/rat.sh
rat-excludes 
test/python/cuda_helper.py
test/python/test_memoryPool.py

The other files in PR are not modified for this PR, but they are added by git due to the merge.

nudles:
I think a better way is to 
1. checkout from the latest dev
2. copy the updated files from the master to the dev
3. commit.

Otherwise, the commits would be mixed.
Thanks!

PR #625, opened 2020-03-11
tensor.AsType()

dcslin:


dcslin:
Hi @joddiy , before i clean the code, please check if this works for you?
https://github.com/apache/singa/blob/41eca964faf7de14cea53c6ba465ce44e7cdfaa4/test/python/test_tensor.py#L340

joddiy:
Hi, shicong, it works fine, thanks.

src/core/tensor/math_kernel.cu
```diff
@@ -184,6 +184,20 @@ __global__ void KernelAbs(const size_t n, const float *in, float *out) {
   }
 }
 
+__global__ void KernelAsType2(const size_t n, const float *in, int *out) {
```
nudles: can you give a better name?
like KernelCastInt2Float or KernelInt2Float

src/core/tensor/math_kernel.cu
```diff
@@ -184,6 +184,20 @@ __global__ void KernelAbs(const size_t n, const float *in, float *out) {
...
+  for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < n;
+       i += blockDim.x * gridDim.x) {
+    out[i] = in[i];
```
nudles: do you need the cast operation like `out[i]=int(in[i])?

dcslin: I think it is working as implicit casting now

src/core/tensor/math_kernel.cu
dcslin: Thank you, this is rectified.

dcslin:
This is ready for merge

src/core/tensor/math_kernel.cu
nudles: better make it explicit.

dcslin: thank you for the suggestion. it is explicited now.

PR #628, opened 2020-03-14
Update conda recipe to use conda dnnl

dcslin:


chrishkchris:
seems great! only the macos version cannot find dnnl?

chrishkchris:
the travis test results still cannot find "deprecated"
I think the "deprecated" should be put at "run" rather than "host" in conda_build_config.yaml, because it is used at runtime (while build process doesn't need).

dcslin:
> the travis test results still cannot find "deprecated"
> I think the "deprecated" should be put at "run" rather than "host" in conda_build_config.yaml, because it is used at runtime (while build process doesn't need).

@chrishkchris thanks for the advice

dcslin:
Should be ok now.

Issue #630, opened 2020-03-17
AsType cannot work when after Reshape


joddiy:
Hi, @dcslin , The asType cannot work when it is after the reshape operators.

Please check by using the following code:

```
    dev = device.create_cuda_gpu()
    X = np.array([[1, 0], [1, 1]]).astype(np.int32)
    x = tensor.from_numpy(X)
    x.to_device(dev)

    x = autograd.cast(x, tensor.int32)
    x = autograd.reshape(x, [1, 2, 2])
    x = autograd.cast(x, tensor.float32)
```

dcslin:
hi @joddiy ,thank you for raising the issue, it should be working after this pr
https://github.com/apache/singa/blob/205a2832b23694b748134ef69ef9c0b4c642afd5/test/python/test_tensor.py#L341

joddiy:
Thans shicong, it works fine for me.

PR #632, opened 2020-03-19
make astype return new tensor

dcslin:


dcslin:
fix: https://github.com/apache/singa/issues/630

src/core/tensor/tensor.cc
```diff
@@ -152,29 +152,24 @@ Tensor Resize(const Tensor &in, const Shape &shape) {
     }                                                                          \
   } while (0)
 
-Tensor &Tensor::AsType(const DataType type) {
+// return new tensor
+Tensor Tensor::AsType(const DataType type) {
```
nudles: do we need to add a flag for in-place? default can be False.

dcslin: Yes. We could add a flag to indicate this explictly. But second thought is, by adding in-place flag, `in-place=True` is also expected to be available. Then, it go back to the previous discussion, if we want to do `in-place` for multi-referenced tensor, what is the expected behavior?

nudles: Do we have `in-place` for any other tensor functions?
the behavior is 
1. in-place = true, then reuse the memory (only works if the bitwidth is the same for the two data types).
2. in-place= false, return a tensor with a new memory block.

case 1 is a bit confusing for the case that the bitwidths are different for the two data types although it may save some cost.

so let's just ignore the in-place flag.

is your concern in the same?

dcslin: Yes. this is my concern.

Currently the `in place` flag are available in python tensor api. No one in cpp tensor api yet.

Issue #634, opened 2020-03-24
cannot do matmul for high-dim tensors


joddiy:
Hi, @dcslin , we ccannot do matmul for high-dim tensors:

The error is:
> F0324 05:47:19.174587 15611 tensor.cc:1413] Check failed: A.shape().size() == 2u (4 vs. 2)

please use this test case:
```python3
x1 = np.random.randn(1, 12, 256, 64).astype(np.float32)
x2 = np.random.randn(1, 12, 64, 256).astype(np.float32)
x1 = tensor.from_numpy(x1)
x1.to_device(gpu_dev)
x2 = tensor.from_numpy(x2)
x2.to_device(gpu_dev)

y = autograd.Matmul()(x1, x2)
print(tensor.to_numpy(y[0]))
```

dcslin:
Hi @joddiy , for tensor mul >2d, please use `tensor.tensordot`, it depends on which axis to perform the `tensordot`
```python
    def test_4d_tensor_dot(self):
        for dev in [cpu_dev, gpu_dev]:
            x1 = np.random.randn(1, 12, 256, 64).astype(np.float32)
            x2 = np.random.randn(1, 12, 64, 256).astype(np.float32)
            x1 = tensor.from_numpy(x1)
            x1.to_device(dev)
            x2 = tensor.from_numpy(x2)
            x2.to_device(dev)

            y = tensor.tensordot(x1, x2, axes=([3],[2]))
            print(y.shape)

            y = tensor.tensordot(x1, x2, axes=([2,3],[3,2]))
            print(y.shape)

```
output:
```
(1, 12, 256, 1, 12, 256)
(1, 12, 1, 12)
(1, 12, 256, 1, 12, 256)
(1, 12, 1, 12)
...........................
----------------------------------------------------------------------
Ran 27 tests in 0.121s

OK
```

reference: https://docs.scipy.org/doc/numpy/reference/generated/numpy.tensordot.html

joddiy:
Hi, shicong, according to the np's definition, the matmul for high-dim matrix should be:

```python3
x1 = np.random.randn(1, 12, 256, 64).astype(np.float32)
x2 = np.random.randn(1, 12, 64, 256).astype(np.float32)
print(np.matmul(x1, x2).shape)
```

Output:
> (1, 12, 256, 256)

Actually, it does the matmul for (256, 64) * (64, 256) for all cells of the first axes (1, 12).

dcslin:
[...]

PR #639, opened 2020-03-26
added support for matmul 3d 4d

dcslin:


dcslin:
fix issue: https://github.com/apache/singa/issues/634

dcslin:
fix https://github.com/apache/singa/issues/634 follow up issue

nudles:
is matmul different to tensordot?
why not call tensordot to implement it?

nudles:
Is it called batched matmul or batched gemm?
Seems different to tensordot.
since the API for matmul remains the same, we may need to update the DocString.

dcslin:
> is matmul different to tensordot?
> why not call tensordot to implement it?

Batch matmul could not be implemented by tensordot, as matmul is not really tensor dot.
Batch matmul could be illustrated by einsum:
``` python
>>> x1.shape
(2, 3, 4, 5)
>>> x2.shape
(2, 3, 5, 6)
>>> np.testing.assert_almost_equal(np.matmul(x1,x2), np.einsum('ijkl,ijlm->ijkm',x1,x2))
# output is (2,3,4,6)
```


Boardcasting matmul could be implemented by tensordot:
```
>>> x1.shape
(2, 3, 4, 5)
>>> x3.shape
(5, 6)
>>> np.testing.assert_almost_equal(np.matmul(x1,x3), np.tensordot(x1,x3,axes=([3],[0])))
# output is (2,3,4,6)
```

dcslin:
> Is it called batched matmul or batched gemm?
> Seems different to tensordot.
> since the API for matmul remains the same, we may need to update the DocString.

In the high level API, `batch matmul` is ok, I see people calling it this way. In numpy, all `matmul` are call `matmul`.

In the low level API, cublas calls it `Batched GEMM`, cblas also does similarly.

I have update `singa/tensor.py` docstring to cover 3 cases:
1. 2d*1d matrix to vector
2. 2d*2d matrix to matrix
3. 3/4d*3/4d batched matrix to matrix

PR #640, opened 2020-03-27
Add onnx in dockerfile

chrishkchris:


nudles:
seems some libs are not compatible.

chrishkchris:
> seems some libs are not compatible.

Since the new PR #639 based on the new conda package version protobuf 3.9.2 can pass the travis build test, this PR without the conda-forge channel cannot pass. 
So there is still some other library needed to copy from conda-forge to nusdbsystem.
I will need to think which one is missing

chrishkchris:
I changed this PR to "add onnx in dockerfile" to support developers

PR #643, opened 2020-04-01
Append public key of wangwei into KEY

nudles:


nudles:
@moazreyad  could you check this PR?

PR #645, opened 2020-04-01
add function comments for autograd

joddiy:


python/singa/autograd.py
```diff
@@ -631,16 +712,39 @@ def backward(self, dy):
 
 
 def add_bias(x, b, axis=0):
+    """
+    Add Bias to each row / column of the Tensor, depending on the axis arg.
+    Args:
+        x: matrix.
+        b (CTensor): bias to be added.
```
nudles, line 719: pls try to make the docstring consistent.
e.g., some variables have (CTensor) whereas some do not have.

python/singa/autograd.py
```diff
@@ -631,16 +712,39 @@ def backward(self, dy):
...
-
 class Reshape(Operation):
+    """
+    Reshape the input tensor similar to numpy.reshape. 
+    """
 
     def __init__(self, shape):
+        """
+        Args:
+            shape: Specified shape for output. At most one dimension of the new shape can be -1. In this case,
```
nudles, line 734: try to keep  the length of a line within 80.

python/singa/autograd.py
```diff
@@ -697,15 +833,32 @@ def backward(self, dy):
...
+    Performs element-wise binary addition.
+    """
 
     def __init__(self):
         super(Add, self).__init__()
 
     def forward(self, a, b):
+        """
+        Args:
+            a (CTensor): First operand..
```
nudles, line 857: For simple functions, we can simplify the docstring, e.g., no need to use "First operand.." or "Second operand". You can put "Return a+b, where a and b are CTensor."

python/singa/autograd.py
```diff
@@ -725,12 +885,26 @@ def backward(self, dy):
 
 
 def add(a, b):
+    """
+    Performs element-wise binary addition.
+    Args:
+        a (CTensor): First operand..
```
nudles, line 891: a is Tensor or CTensor?

nudles:
docString is to explain the APIs.
We do not need to make it wordy, just make it clear. 
E.g., for `a + b`, we do not need to have "a" is the first operand. Instead, it is more useful to explain if a and b must be of the same shape, is boradcast done automatically.

python/singa/autograd.py
joddiy: fixed, and I have checked the whole file again.

python/singa/autograd.py
joddiy: fixed, and I have checked the whole file again.

python/singa/autograd.py
joddiy: fixed, and for other simple operators, such as or, and, etc, I also change them like this.

python/singa/autograd.py
joddiy: sorry, I made a mistake, it should be Tensor, I've checked all the file again to fix it.

joddiy:
@nudles  I've checked the whole file again and fixed all the problems. I will create another PR later.

PR #646, opened 2020-04-01
SINGA-487 Test case for distributed module

chrishkchris:


chrishkchris:
To make the test case complete, a test case is added for distributed module, which can run using one gpu as well as using mpich to run with multiple gpus

chrishkchris:
ready for merge

chrishkchris:
I will add more test case first

chrishkchris:
Added more test cases, ready for merge

chrishkchris:
@XJDKC I also added `self.fused_all_reduce([p.data], send=False)` in the function `backward_and_partial_update`
to adopt the change last time

chrishkchris:
> @XJDKC I also added `self.fused_all_reduce([p.data], send=False)` in the function `backward_and_partial_update`
> to adopt the change last time in the `fused_all_reduce` function

but I think carefully cannot modify like this, the logic might need to rewrite

chrishkchris:
ready for merge

XJDKC:
> @XJDKC I also added `self.fused_all_reduce([p.data], send=False)` in the function `backward_and_partial_update`
> to adopt the change last time in the `fused_all_reduce` function

You can pass the whole plist to the fused_all_reduce function, no need to pass tensors one by one.

chrishkchris:
> > @XJDKC I also added `self.fused_all_reduce([p.data], send=False)` in the function `backward_and_partial_update`
> > to adopt the change last time in the `fused_all_reduce` function
> 
> You can pass the whole plist to the fused_all_reduce function, no need to pass tensors one by one.

done, it becomes
```
                        self.fused_all_reduce(plist, send=False)
                        self.fused_all_reduce(plist)
```

chrishkchris:
> > > @XJDKC I also added `self.fused_all_reduce([p.data], send=False)` in the function `backward_and_partial_update`
> > > to adopt the change last time in the `fused_all_reduce` function
> > 
> > 
> > You can pass the whole plist to the fused_all_reduce function, no need to pass tensors one by one.
> 
> done, it becomes
> 
> ```
>                         self.fused_all_reduce(plist, send=False)
>                         self.fused_all_reduce(plist)
> ```




The asychronous training is also okay now:
```
root@71ac539cda77:~/dcsysh/singa/examples/autograd# python3 cifar10_multiprocess.py
Starting Epoch 0:
Training loss = 2910.528564, training accuracy = 0.335027
Evaluation accuracy = 0.438902, Elapsed Time = 148.023807s
Starting Epoch 1:
Training loss = 2098.108887, training accuracy = 0.512984
Evaluation accuracy = 0.558393, Elapsed Time = 147.847368s
Starting Epoch 2:
Training loss = 1696.253052, training accuracy = 0.612596
Evaluation accuracy = 0.634615, Elapsed Time = 145.889916s
Starting Epoch 3:
Training loss = 1424.486328, training accuracy = 0.678877
Evaluation accuracy = 0.725461, Elapsed Time = 144.518893s
Starting Epoch 4:
Training loss = 1202.804688, training accuracy = 0.732714
Evaluation accuracy = 0.731070, Elapsed Time = 145.818738s

```

chrishkchris:
Ready for merge

nudles:
Does the test case need >=2 GPUs?
If so, we need to disable it, if the environment only has 0 or 1 GPU.

chrishkchris:
> Does the test case need >=2 GPUs?
> If so, we need to disable it, if the environment only has 0 or 1 GPU.

The code works when there is at least 1 GPU. It supports two mode to run:
(i) Run the code directly, e.g. `python test_dist.py `
(ii) Run with mpich, e.g. `mpiexec -np 2 test_dist.py` for 2 GPUs
In the first case, there is only one rank (i.e. rank 0), while it can test if the communicator contructor and the rountine to convert 32 to 16 bits, sparsifciation, etc, are correct or not.

Without GPU, the test code cannot run but it does not affect the Travis CI build and test process, Other GPU test in test_operation.py has the same problem 
we can see:
https://travis-ci.org/github/apache/singa/jobs/669989104#L4012 Cannot find gpu in dist_test.py
https://travis-ci.org/github/apache/singa/jobs/669989104#L4028 Cannot find onnx
https://travis-ci.org/github/apache/singa/jobs/669989104#L4078 Cannot find gpu in test_operation.py

PR #647, opened 2020-04-01
update tensor.py docstring according to google docstring style

dcslin:


dcslin:
used this tool to lint the docstring according to google docstring style:
https://github.com/terrencepreilly/darglint

PR #652, opened 2020-04-04
fix unable to build error

dcslin:


dcslin:
not sure why, but using latest miniconda can fix various errors when building the images

dcslin:
all 5 docker files are tested and pushed to https://hub.docker.com/r/slin004/singa/tags
* devel:
3.0.0-cpu-devel-ubuntu18.04
3.0.0-cuda9.0-cudnn7.4.2-devel-ubuntu16.04
3.0.0-cuda10.0-cudnn7.4.2-devel-ubuntu18.04

* runtime:
3.0.0-cuda9.0-cudnn7.4.2-ubuntu16.04
3.0.0-cpu-ubuntu16.04

tool/docker/runtime/cpu/Dockerfile
```diff
@@ -28,10 +28,12 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # install conda
-RUN wget --no-check-certificate https://repo.continuum.io/miniconda/Miniconda3-4.5.12-Linux-x86_64.sh -O miniconda.sh;
+RUN wget --no-check-certificate https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
```
nudles, line 31: is it better to fix the version of miniconda?
if miniconda has a new version, is it possible that singa will fail to build/install?

dcslin: yes. thanks. this is updated.

PR #653, opened 2020-04-04
conda build fix git get tag

dcslin:


dcslin:
fix git get tag and all tested as:
https://anaconda.org/shicong/singa
https://anaconda.org/shicong/singa-cpu
https://anaconda.org/shicong/singa-gpu

dcslin:
fixed 2 bug:
1. need to add conda `source` to `gpu/meta` and `cpu/meta`
2. conda dont accept `-` in the version. like `3.0.0-rc0`. Thus, replace `-` with `.` and make the example to be `3.0.0.rc0`.

PR #659, opened 2020-04-07
Improve graph compatibility

XJDKC:
Capture some tensors by value. These tensors will be recycled during the evaluation phase. But this PR may affect memory usage when enabling graph.

### DIFF
--- a/python/singa/device.py
+++ b/python/singa/device.py
@@ -97,14 +97,17 @@ def create_cuda_gpus(num):
     return singa.Platform.CreateCudaGPUs(num)
 
 
-def create_cuda_gpu():
+def create_cuda_gpu(set_default=True):
     '''Create a single CudaGPU device.
 
     Returns:
         a swig converted CudaGPU device.
     '''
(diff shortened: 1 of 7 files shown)


XJDKC:
Many intermediate tensors are recorded which may affect the performance of automatic memory recycling. The scheduler won't recycle the tensor used on the Python side.
[example](https://github.com/apache/singa/blob/dev/python/singa/autograd.py#L2013)
x, mean and var are recorded on the Python side and they always exist after every iteration, so the scheduler won't recycle them, but they could have been recycled.

nudles:
why do we need to capture some tensors by value and some by reference?

XJDKC:
> why do we need to capture some tensors by value and some by reference?

This is exactly what I did before. But for the evaluation phase, tensors in the example will be recycled(It will create a new operation class by calling __call__ function, so the previous operations will be recycled and so do tensors they contain). If we access them in the next epoch, the program will crash.

XJDKC:
I have fixed this issue, now the peak memory usage is even lower than the previous test. It probably saves another 0.5GB when training ResNet50.

PR #660, opened 2020-04-07
Update test cases for both cpu and gpu

joddiy:


test/python/test_api.py
```diff
@@ -201,43 +201,47 @@ def _run_testing(x_0, s_0, b_0, rm_0, rv_0, m_0=0.1):
...
+                x_0 = x_0.reshape(aft_shape)
+                x_0 = x_0 - np.max(x_0)
+                y1 = np.divide(np.exp(x_0),
+                               np.sum(np.exp(x_0), axis=1).reshape(x_0.shape[0],
+                                                                   1))  # 2d softmax
+                y1 = y1.reshape(org_shape)
+
+                np.testing.assert_array_almost_equal(tensor.to_numpy(y0), y1)
+
+                _run_test(dev, [2, 2], 1, [2, 2])
```
nudles, line 223: indent error?

test/python/test_api.py
```diff
@@ -201,43 +201,47 @@ def _run_testing(x_0, s_0, b_0, rm_0, rv_0, m_0=0.1):
 
     def test_softmax_api(self):
```
nudles, line 202: in the future, we can improve it as follows
```python
def softmax_api_test(dev)

def test_softmax_api_cpu(self)
   cpu = ...
   softmax_api_test(cpu)

@unittestskip(...)
def test_softmax_api_gpu(self)
   gpu = ...
   softmax_api_test(gpu)
```
In this way, the unittest will report if the error is from cpu test or gpu test. Otherwise, it will only report that the softmax_api has error.

joddiy: fixed

test/python/test_api.py
joddiy: fixed

PR #662, opened 2020-04-07
CUDNN LSTM

dcslin:


dcslin:
todo:
[ ] add lstm
[ ] read `autograd.py` rnn and make compatible

nudles:
I think they are there for future extension.

dcslin:
> I think they are there for future extension.

ok thanks.

Currently, fixing something related to cuda `column major order`, input are stored in `{seq length, batch size, feature dim}` manner. And `batch size` might vary.

dcslin:
Hi @nudles , I am confused by quite a few things here and there. Primarily is the input layout. May I have your comment on the input layout please?
Does autograd.py RNN expect input layout to be `list(Tensor, ...)`, where `len(list) == seq length/time steps`, and for each tensor, tensor shape is `batch size, input dim`?

nudles:
we can just follow cudnn format and update autograd lstm.

dcslin:
It is basically follow autograd LSTM structure, with additional arguments added. 
And the data input and output are following previous convention, which is also the cudnn convention. Input are in a array of {batch size, feature dim} with length - seq length/time steps.
I have not edit autograd.py before, kindly advise anything wrong. Thanks

python/singa/autograd.py
```diff
@@ -3239,95 +3239,176 @@ def __init__(
...
-        self.device_check(inputs[0], *(self.Wx + self.Wh + self.Bx + self.Bh))
-        batchsize = xs[0].shape[0]
-        out = []
-        h, c = self.step_forward(xs[0], h0, c0, self.Wx, self.Wh, self.Bx,
-                                 self.Bh)
-        out.append(h)
-        for x in xs[1:]:
-            assert x.shape[0] == batchsize
-            h, c = self.step_forward(x, h, c, self.Wx, self.Wh, self.Bx,
+        if self.backend == "singa":
```
nudles, line 3301: LSTM is a Layer subclass.
It should call lstm function to do the forward by passing the parameters to it.
lstm function creates _LSTM operation, which implements the forward() and backward().

Pls check the _Conv2d(Operation), conv2d() function and Conv2d(Layer).

nudles: I think it's better to separate the existing LSTM with the cudnn RNN. It may get you confused.
For the cudnnRNN Layer, it has a type, {vanilla, lstm, gru} to configure.

python/singa/autograd.py
```diff
@@ -3239,95 +3239,176 @@ def __init__(
...
+                b = Tensor(shape=Bx_shape, requires_grad=True, stores_grad=True)
+                b.set_value(0.0)
+                self.Bh.append(b)
+
+            self.params = self.Wx + self.Wh + self.Bx + self.Bh
+        elif backend == "cudnn":
+            if not singa.USE_CUDA:
+                raise Exception("Could not use cudnn without cuda compiled.\n")
+            if not inputs:
+                raise Exception("Input is required for init cudnn LSTM.\n")
```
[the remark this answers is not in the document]
dcslin: this is similar to c++ operation conv handle constructor that requires input for constructing cudnn desc.

python/singa/autograd.py
```diff
@@ -3239,95 +3239,176 @@ def __init__(
...
+
+            CUDNN_LSTM_MODE = 2
```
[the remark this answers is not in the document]
dcslin: This is added.

python/singa/autograd.py
dcslin: thank you for the explanation, this is changed to layer-operation model.

python/singa/autograd.py
```diff
@@ -3239,176 +3238,152 @@ def __init__(self,
...
+        cout1 = mul(f, c)
+        cout2 = mul(i, g)
+        cout = add(cout1, cout2)
+
+        hout = tanh(cout)
+        hout = mul(o, hout)
+        return hout, cout
+
+
+class _RNN_cudnn(Operation):
```
nudles, line 3333: rename to _RNN?

python/singa/autograd.py
```diff
@@ -3239,176 +3238,152 @@ def __init__(self,
...
+    def __init__(self, handle):
+        super(_RNN_cudnn, self).__init__()
```
nudles, line 3335: assert cuda is available.

nudles: add TODO for cpu case.

python/singa/autograd.py
```diff
@@ -3239,176 +3238,152 @@ def __init__(self,
...
+        self.handle = handle
 
-            self.W = Tensor(shape=(self.rnn_handle.weights_size,),
-                            requires_grad=True,
-                            stores_grad=True)
-            self.W.gaussian(0.0, 1.0)
+    def forward(self, x, W):
+        y = singa.GpuRNNForwardTraining(x, W, self.handle)
```
nudles, line 3339: if training is false, need to call GpuRNNForwardInference?

python/singa/autograd.py
```diff
@@ -3239,176 +3238,152 @@ def __init__(self,
...
-            if training:
-                self.buffer = {"cpp_y": cpp_y, "cpp_x": cpp_x}
+        dx = singa.GpuRNNBackwardx(self.inputs[2], dy, self.inputs[1], self.handle)
+        dW = singa.GpuRNNBackwardW(self.inputs[0], self.inputs[2], self.handle)
+        return dx, dW
 
-            return y
-        else:
-            raise Exception("Unsupported backend %s.\n" % backend)
+class RNN_cudnn(Layer):
```
nudles, line 3352: Rename to RNN.

python/singa/autograd.py
```diff
@@ -3239,176 +3238,152 @@ def __init__(self,
...
+    def __init__(self, input_size, hidden_size, rnn_mode=2):
+        """
+            Args:
+                input_size: input feature dim
+                hidden_size: hidden feature dim
+                rnn_mode: could be 0 - RNN RELU, 1 - RNN TANH, 2 - LSTM, 3 - GRU
```
nudles: use string, 'relu', 'tanh', gru'.
does it support vanilla rnn with 'tanh' ?

python/singa/autograd.py
```diff
@@ -3239,176 +3238,152 @@ def __init__(self,
...
-    def backward(self, dy):
-        if self.backend == "cudnn":
-            if not singa.USE_CUDA:
-                raise Exception("Could not use cudnn without cuda compiled.\n")
+    def __call__(self, x):
+        if not hasattr(self, "handle"):
+            cpp_x = singa.VecTensor()
+            [cpp_x.append(i.data) for i in x]
+
+            self.handle = singa.CudnnRNNHandle(cpp_x, self.input_size, self.hidden_size, self.rnn_mode)
```
nudles: assert gpu is available.

dcslin: this is resolved

python/singa/autograd.py
dcslin: this is resolved

python/singa/autograd.py
dcslin: `RNN` is used for existing highlevel implementation
After consideration, `RNN_direct` is used to name newly added C++ backed (cudnn/cpu) RNN

python/singa/autograd.py
dcslin: this is resolved

python/singa/autograd.py
dcslin: this is resolved

python/singa/autograd.py
dcslin: this is resolved

src/api/model_operation.i
```diff
@@ -188,6 +189,50 @@ Tensor GpuPoolingForward(const CudnnPoolingHandle &cph, const Tensor &x);
...
+class CudnnRNNHandle {
+ public:
+   CudnnRNNHandle(const std::vector<Tensor> &x,
+       const int feature_size,
+       const int hidden_size,
+       const int mode = 0,
+       const int num_layers = 1,
+       const int  bias = 1,
+       const float dropout = 0.0f,
+       const int bidirectional = 0);
```
nudles, line 202: num_layers, dropout and bidirectional can be exposed to python?

python/singa/autograd.py
```diff
@@ -3330,6 +3330,94 @@ def step_forward(self, x, h, c, Wx, Wh, Bx, Bh):
...
+            self, "inputs"), "Please set training as True before do BP. "
+
+        # TODO: CPU backward
+
+        # GPU backward
+        dx = singa.GpuRNNBackwardx(self.inputs[2], dy, self.inputs[1], self.handle)
+        dW = singa.GpuRNNBackwardW(self.inputs[0], self.inputs[2], self.handle)
+        return dx, dW
+
+class RNN_direct(Layer):
```
nudles, line 3364: rename to RNN?

dcslin: RNN is used for existing highlevel implementation
After consideration, RNN_direct is used to name newly added C++ backed (cudnn/cpu) RNN

python/singa/autograd.py
```diff
@@ -3330,6 +3330,94 @@ def step_forward(self, x, h, c, Wx, Wh, Bx, Bh):
...
+
+    def __call__(self, x):
+        if not hasattr(self, "handle"):
+            cpp_x = singa.VecTensor()
+            [cpp_x.append(i.data) for i in x]
+
+            # TODO: CPU handle
+
+            # GPU handle
+            self.handle = singa.CudnnRNNHandle(cpp_x, self.input_size, self.hidden_size, self.cudnn_rnn_mode)
```
nudles, line 3404: can infer the input_size from x?
then there is not need to require input_size in the constructor.

python/singa/autograd.py
```diff
@@ -3330,6 +3330,94 @@ def step_forward(self, x, h, c, Wx, Wh, Bx, Bh):
...
+
+            self.W = Tensor(shape=(self.handle.weights_size,),
+                            requires_grad=True,
+                            stores_grad=True)
+            self.W.gaussian(0.0, 1.0)
+
+        return _RNN(self.handle)(x, self.W)[0]
```
nudles, line 3411: rnn models may return a list of tensors.
therefore, we should not only select the first.
but need to add some comments to interpret the returned list.
for different modes, the returned list could be different.

python/singa/autograd.py
```diff
@@ -3330,6 +3330,94 @@ def step_forward(self, x, h, c, Wx, Wh, Bx, Bh):
...
+        if not hasattr(self, "handle"):
+            cpp_x = singa.VecTensor()
+            [cpp_x.append(i.data) for i in x]
+
+            # TODO: CPU handle
+
+            # GPU handle
+            self.handle = singa.CudnnRNNHandle(cpp_x, self.input_size, self.hidden_size, self.cudnn_rnn_mode)
+
+            self.W = Tensor(shape=(self.handle.weights_size,),
```
nudles, line 3406: if the layer parameters are set outside (e.g., after loading the checkpoint file) via set_params, there is no need to randomly initialize W.

python/singa/autograd.py
```diff
@@ -3330,6 +3330,94 @@ def step_forward(self, x, h, c, Wx, Wh, Bx, Bh):
...
+                            requires_grad=True,
+                            stores_grad=True)
+            self.W.gaussian(0.0, 1.0)
+
+        return _RNN(self.handle)(x, self.W)[0]
+
+    def get_params(self):
+        return self.W
```
nudles, line 3414: since W is created after the layer is called in forward propagation, if this function is called after the layer is created, there will be error.
we can add an assert to print some hints, e.g., W is not initialized until the layer is forwarded with data for at least once

python/singa/autograd.py
nudles: let's unify all the rnn implementation in this way

```python
def RNN(hidden_size, num_layers=1, dropout=0.0, direction='one', return_seq=False, return_state=False, alg='lstm', backend='cudnn') :
   if backend == 'cudnn':
      reutrn CudnnRNN(hidden_size, num_layers, dropout, direction, mode)
   if backend == 'native':
     if mode == 'lstm':
      return LSTM()
     elif mode == 'vanilla':
       return VanillaRNN()
     ....
```

All rnn classes inherit 
```python
class RNNBase(Layer):
   def __init__(self, hidden_size, num_layers=1, dropout=0.0, direction='one', return_seq=False, return_state=False, alg='lstm')
```

python/singa/autograd.py
dcslin: In the cudnn settings, x is a list of {batchsize, featuresize} with length `seq_length`, and batchsize might vary. And also there are `seq_length` of `CudnnTensorDescriptor` which can have different batch size.

We might allow only fixed value of batch size in one batch of input. Then we could remove `x` from constructor.

Side note 1. current Conv2d is using similar design, `x` in constructor
2. native rnn implementation currently has similar API, x is a list of {batchsize, featuresize} with length `seq_length`

python/singa/autograd.py
dcslin: this is rectified

src/api/model_operation.i
dcslin: this is rectified

python/singa/autograd.py
dcslin: this is rectified

python/singa/autograd.py
dcslin: this is rectified

python/singa/autograd.py
dcslin: this is rectified

python/singa/autograd.py
dcslin: mentioned in below review

PR #667, opened 2020-04-08
Fix bugs of training

XJDKC:
* Fix bugs of buffering Mult.
* Can train ResNet50 with batch size of 64 now.

### DIFF
--- a/src/core/tensor/tensor.cc
+++ b/src/core/tensor/tensor.cc
@@ -161,6 +161,9 @@ Tensor Resize(const Tensor &in, const Shape &shape) {
 
 // return new tensor
 Tensor Tensor::AsType(const DataType type) {
+  CHECK(block() && block()->initialized() == true)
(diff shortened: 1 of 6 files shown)


XJDKC:
There is another commit to be pushed.

src/core/tensor/tensor.cc
```diff
@@ -161,6 +161,9 @@ Tensor Resize(const Tensor &in, const Shape &shape) {
 
 // return new tensor
 Tensor Tensor::AsType(const DataType type) {
+  CHECK(block() && block()->initialized() == true)
+      << "the data of the tensor needs be initialized before casting to "
```
nudles, line 165: this cannot fix the error. need to update the test case to reverse the two lines.

XJDKC: Done.

XJDKC: There seem to be many such cases in unit testing

PR #684, opened 2020-04-21
ONNX supports module

joddiy:
This PR adds a class to create a [class](https://github.com/apache/singa/pull/684/files#diff-4839d7350844248bf25a18751ed06062R2087-R2161) inheriting from the autograd.module class, and now the ONNX model can be trained as SINGA model.

image

However, this PR supports most ONNX models except BERT, since BERT uses lots of operators which don't have a gradient, may need to discuss it further.

TO-DO List:
- ~~add base class supports module.~~
- add training examples for:
    * ~~MobileNet~~
    * ResNet-18, 34, 50, 101, 152, v1 for pre-activation, v2 for post-activation
    * SqueezeNet
    * VGG-16, 19
    * AlexNet
    * GoogleNet
    * RCNN_ILSVRC13
    * DenseNet
    * Inception_V1
    * Inception_V2
    * ShuffleNet_V1
    * ShuffleNet_V2
    * ZFNet-512

### DIFF
--- a/python/singa/sonnx.py
+++ b/python/singa/sonnx.py
@@ -2094,7 +2094,7 @@ class BaseModel(module.Module):
         """
         super(BaseModel, self).__init__()
         self.optimizer = optimizer
-        self.loss = loss
+        self.ls = loss
         self.g = sg_ir.model.graph
         self.tm = sg_ir.tensor_map
         self.oq = sg_ir.singa_ops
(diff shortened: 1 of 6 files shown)


python/singa/sonnx.py
```diff
@@ -2135,6 +2084,81 @@ def run(self, inputs, **kwargs):
...
+
+    def __init__(self, sg_ir, loss, optimizer):
+        """
+        Init a SIGNA Module
+        Args:
+            sg_ir: a SingaRep class
+        """
+        super(BaseModel, self).__init__()
+        self.optimizer = optimizer
+        self.loss = loss
```
XJDKC, line 2097: Please change the name of this member variable. Since loss is already the name of a member function, this assignment will rewrite the original function. So the loss function is not wrapped by the decorator(buffer_operation) which will cause a new operation to be created in each iteration.

Issue #681, opened 2020-04-20
AlexNet bacward shape missmatch + ReLu return a tuple


Belegkarnil:
Hi,

I have implemented AlexNet in singa but I obtain an error during the backward_and_update instruction. I am using Singa 3.0.0.rc1 on cpu.

This is my AlexNet implementation:
`from singa import autograd
from singa import module
from singa import opt

__all__ = ['AlexNet', 'alexnet']

class AlexNet(module.Module):
	def __init__(self, num_classes=1000):
		super(AlexNet, self).__init__()
		# 12 sur GPU donc 6 & 6
		self.features1 = [
			autograd.Conv2d(3,64,kernel_size=11,stride=4,padding=2),
			autograd.ReLU(),
			autograd.MaxPool2d(kernel_size=3, stride=2),
			autograd.Conv2d(64,192,kernel_size=5,padding=2),
			autograd.ReLU(),
			autograd.MaxPool2d(kernel_size=3, stride=2),
			autograd.Conv2d(192,384,kernel_size=3,padding=1),
			autograd.ReLU(),
			autograd.Conv2d(384, 256,kernel_size=3,padding=1),
			autograd.ReLU()
		]
		self.features2 = [
			autograd.Conv2d(256, 256,kernel_size=3,padding=1),
			autograd.ReLU(),
			autograd.MaxPool2d(kernel_size=3, stride=2)
		]
		self.avgpool = autograd.AvgPool2d(6, stride=1)
		self.flatten = autograd.Flatten()
		self.classifier = [
			autograd.Dropout(),
			autograd.Linear(256 * 6 * 6, 4096),
			autograd.ReLU(),
			autograd.Dropout(),
			autograd.Linear(4096, 4096),
			autograd.ReLU(),
			autograd.Linear(4096, num_classes)
		]
		self.optimizer = opt.SGD(lr=0.001, momentum=0.9)
	def loss(self, out, ty):
		return autograd.softmax_cross_entropy(out, ty)
	def optim(self, loss, dist_option, spars):
		if dist_option == 'fp32':
			self.optimizer.backward_and_update(loss)
		elif dist_option == 'fp16':
			self.optimizer.backward_and_update_half(loss)
		elif dist_option == 'partialUpdate':
			self.optimizer.backward_and_partial_update(loss)
		elif dist_option == 'sparseTopK':
			self.optimizer.backward_and_sparse_update(loss, topK=True, spars=spars)
		elif dist_option == 'sparseThreshold':
[...]

PR #685, opened 2020-04-22
Add Alexnet Example

dcslin:


chrishkchris:
Here is a test on alexnet using cifar10:

```
(singapy36) dcsysh@panda11:~/singa/examples/cnn$ python train.py alexnet cifar10
Starting Epoch 0:
Training loss = 1611.267456, training accuracy = 0.225992
Evaluation accuracy = 0.307091, Elapsed Time = 69.078020s
Starting Epoch 1:
Training loss = 1349.916016, training accuracy = 0.356514
Evaluation accuracy = 0.385917, Elapsed Time = 69.427504s
Starting Epoch 2:
Training loss = 1189.298340, training accuracy = 0.437600
Evaluation accuracy = 0.456430, Elapsed Time = 69.476516s
```

chrishkchris:
One point to notice is that:
Currently in SINGA, ReLU is stateless, i.e. we don't need to initialize a python object for ReLU before use (and we cannot create an instance of ReLU).

dcslin:
not able to reproduce issue
https://github.com/apache/singa/issues/681

thus tried to implement alexnet and it works fine

maybe we should define issue reporting format to better undertanding these kind of the issues:
example of issue reporting format: 
https://github.com/pytorch/pytorch/issues/new?template=bug-report.md

chrishkchris:
I recommend this one is ready for merge, it is good to have Alexnet for research purpose

chrishkchris:
@dcslin 
I suggest you add an line in Alexnet.py 
# the code is modified from
# https://github.com/pytorch/vision/blob/master/torchvision/models/alexnet.py

Please consider follow this example: https://github.com/apache/singa/blob/master/examples/cnn/model/resnet.py#L21

PR #686, opened 2020-04-28
Linear layer should check if the input channels parameter match inputs

dcslin:


dcslin:
`Linear` layer sliently proceed `forward` with wrong `input channels` which causes unexpected behavior in `backward`

dcslin:
Currently, there is no keras model `compile` equivalent for singa to do sanity check on the model

PR #688, opened 2020-05-02
[stall]Layer auto detect input size

dcslin:


dcslin:
changes:
1. an example of make `class Xxx(Operation)` to private `class _Xxx(Operation)`. Because `class Operation` should be only for internal. For the term `Operation` in the user space, it should be the operation functions `def xxx(x):...`. Also these operation function should be used by the user
2. fix bug for `set_param` when given Tensor as params
3. Modified `Linear` constructor to `(self,out_features, in_features=None, bias=True):`. `out_features` comes first, leaving `in_features` as optional. When `Linear` is constructed with only `out_features`, it's params, `W` and `b`, are not initialized. After `set_params` or `forward`/`__call__`,  it's params, `W` and `b`, are initialized.

nudles:
> 
> 
> changes:
> 
>     1. an example of make `class Xxx(Operation)` to private `class _Xxx(Operation)`. Because `class Operation` should be only for internal. For the term `Operation` in the user space, it should be the operation functions `def xxx(x):...`. Also these operation function should be used by the user
> 
>     2. fix bug for `set_param` when given Tensor as params
> 
>     3. Modified `Linear` constructor to `(self,out_features, in_features=None, bias=True):`. `out_features` comes first, leaving `in_features` as optional. When `Linear` is constructed with only `out_features`, it's params, `W` and `b`, are not initialized. After `set_params` or `forward`/`__call__`,  it's params, `W` and `b`, are initialized.

For the last point, it will break the compatibility.. 
there are two solutions
1. use *args and **kwargs
2. assume the old code passes (in_features, out_features, bias=True) and the new code passes (out_features, bias=True), then we check if in_features is None or not to decide the argument order. 

In V4, we can update the API completely.

dcslin:
> > changes:
> > ```
> > 1. an example of make `class Xxx(Operation)` to private `class _Xxx(Operation)`. Because `class Operation` should be only for internal. For the term `Operation` in the user space, it should be the operation functions `def xxx(x):...`. Also these operation function should be used by the user
> > 
> > 2. fix bug for `set_param` when given Tensor as params
> > 
> > 3. Modified `Linear` constructor to `(self,out_features, in_features=None, bias=True):`. `out_features` comes first, leaving `in_features` as optional. When `Linear` is constructed with only `out_features`, it's params, `W` and `b`, are not initialized. After `set_params` or `forward`/`__call__`,  it's params, `W` and `b`, are initialized.
> > ```
> 
> For the last point, it will break the compatibility..
> there are two solutions
> 
> 1. use *args and **kwargs
> 2. assume the old code passes (in_features, out_features, bias=True) and the new code passes (out_features, bias=True), then we check if in_features is None or not to decide the argument order.
> 
> In V4, we can update the API completely.

ok thanks.

Updated:
changes No.4: modified the `__ini__` to parse `*args and **kwargs` for Linear, RNN, LSTM

nudles:
The following APIs should be backward compatible. Please test.

```python

class Linear(Layer):
     def __init__(self, num_output, *args, bias=True, **kwargs):
          # the following block is for backward compatibility. 
          # the old code will all Linear(2, 3), or (2, 3, False)
          if len(args) > 0:
             num_input = num_output
             num_output = args[0]
          if len(args) > 1:
             bias = args[1]

       self.num_output = num_output
       self.bias = bias

class Conv2d(Layer):
       def __init__(self,
                 out_channels,
                 kernel_size,
                 *args,
                 stride=1,
                 padding=0,
                 dilation=1,
                 group=1,
                 bias=True,
                 pad_mode="NOTSET",
                 **kwargs):
         # the old code create the layer like: Conv2d(8, 16, 3)， or Conv2d(8, 16, 3, stride=1)
         # the following code block is for backward compatibility
         if len(args) >0:
           in_channel=out_channel
           out_channel = kernel
           kernel = args[0]
         if len(args) > 1:
           stride = args[1]
         if len(args) > 2:
           padding = args[2]
 ```

dcslin:
update linear constructor, tested ok

joddiy:
Hi, @dcslin , can I use this PR right now? since I need to let the soonx to support the new autograd  api.

dcslin:
> Hi, @dcslin , can I use this PR right now or which operation I can use now? since I need to let the soonx to support the new autograd api.

Hi @joddiy please refer to https://github.com/apache/singa/pull/697

joddiy:
> > Hi, @dcslin , can I use this PR right now or which operation I can use now? since I need to let the soonx to support the new autograd api.
> 
> Hi @joddiy please refer to #697

Thanks, shicong, and one thing to confirm, the name of operation should be `_ReLU` or `ReLU`?

dcslin:
> > > Hi, @dcslin , can I use this PR right now or which operation I can use now? since I need to let the soonx to support the new autograd api.
> > 
> > 
> > Hi @joddiy please refer to #697
> 
> Thanks, shicong, and one thing to confirm, the name of operation should be `_ReLU` or `ReLU`?

I guess you are building the model? then by convention we use `autograd.relu()`

nudles:
> 
> 
> > > > Hi, @dcslin , can I use this PR right now or which operation I can use now? since I need to let the soonx to support the new autograd api.
> > > 
> > > 
> > > Hi @joddiy please refer to #697
> > 
> > 
> > Thanks, shicong, and one thing to confirm, the name of operation should be `_ReLU` or `ReLU`?
> 
> I guess you are building the model? then by convention we use `autograd.relu()`

I suggest to use layer.ReLU() to avoid mixing operators and layers in constructing the model.
Refer to the first post in https://github.com/apache/singa/issues/696#issue-616336790

dcslin:
some of the work is merged into https://github.com/apache/singa/pull/697
the rest need to be revised as api changed
thus closing this PR

PR #703, opened 2020-05-18
Refactor sonnx, test cases and examples

joddiy:
todo-list:
1. <s>sonnx to suit new API(almost done),</s>
2. <s>sonnx to use new autograd API,</s>
3. <s>test cases,</s>
4. <s>debug test operations, </s>
5. <s>debug sonnx backend test cases, </s>
4. <s>examples </s>
6. <s>add re-training examples</s>
7. (if time is enough) add more example(will move to another PR)


examples/cnn/onnx/train.py
```diff
@@ -0,0 +1,307 @@
...
+    # print content
+    with open(os.path.join(os.path.dirname(__file__), 'models.json')) as json_file:
+        model_config = json.load(json_file)
+        model_config = model_config[model]
+
+    download_model(model_config['url'])
+    onnx_model = onnx.load(os.path.join('/tmp', model_config['path']))
+    onnx_model = update_batch_size(onnx_model, batch_size)
+    sg_ir = sonnx.prepare(onnx_model, device=dev)
+    model = sonnx.create_model(sg_ir, autograd.softmax_cross_entropy, sgd)
```
nudles, line 142: are you going to change to the new SONNXModel API?

joddiy: No, sorry for the confusion, please refer the current code.

python/singa/sonnx.py
```diff
@@ -989,1150 +1035,1051 @@ def from_onnx(args):
...
+    def __init__(self, onnx_model):
+        """
+        Init a SIGNA Model
+        Args:
+            onnx_model (ModelProto): a loaded onnx model
+        """
+        super(SONNXModel, self).__init__()
+        self.sg_ir = prepare(onnx_model)
+        for node, operator in self.sg_ir.layers:
+            self.__dict__[node.name] = operator
```
nudles, line 2070: is each operator here a Layer instance?

joddiy: No, you can check the dict of ` _rename_operators`, for the normal operator, it creates the operator from the `autograd.py`, and for the layer operator, it creates it from `layer.py`.

nudles: if operator is not a Layer instance, then the save_states and load_states cannot work, which works by saving the states of the sub-layers.

joddiy: At [Line 2035-2040](https://github.com/apache/singa/pull/703/files/44796e63f66f1010ac08ff2c7ba8e94160c490c4#diff-4839d7350844248bf25a18751ed06062R2035-R2040), the code will check the type of the instance, if it's from autograd, it will call `setattr(op, key, value)`, otherwise will call `op.set_states(**states)`.

joddiy:
@nudles  ready to merge

examples/onnx/arcface.py
```diff
@@ -78,35 +76,30 @@ def forward(self, x):
...
-    logging.info("preprocessing...")
-    img1, img2 = get_image()
-    img1 = preprocess(img1)
-    img2 = preprocess(img2)
-
-    x_batch = tensor.Tensor(device=dev,
-                            data=np.concatenate((img1, img2), axis=0))
     logging.info("model running...")
-    y = model.forward(x_batch)
+    y = m.forward(*[x])[0]
```
nudles, line 102: why pass `*[x]` to the forward function?
and the [0] should be done within module.forward
```python
def forward():
   y = super(MyModel, self).forward(*x)
   return y[0]
```

joddiy: - the `*[x]` is same with `x`, it'd be ok to change to `x`.

- ok, I'll update the `y[0]`.

PR #708, opened 2020-05-22
SINGA-510 Improvement of Time Profiling Function

chrishkchris:
There are several improvement in the time profiling function
1.  Greatly reduced the measurement overhead: In the new version, we synchronize the cudastream only one time at the end of iteration. The previous version needs to synchronize cudastream whenever it measures each operation, so it introduced time overhead.
2. Can skip the initial iterations for time profiling. It can be set using the function SetSkipIteration, while the default is 5 iterations.
3. Modify the benchmark.py for resnet 50 benchmark time profiling.

### DIFF
--- a/src/core/scheduler/scheduler.cc
+++ b/src/core/scheduler/scheduler.cc
@@ -250,7 +250,7 @@ void Graph::PrintTimeProfiling() {
         // when forward becomes false, it starts the backward propagation
 
         time_elapsed =
-            (nodes_[i]->time_elapsed()) / (iteration_ - skip_iteration_);
+            (nodes_[i]->time_elapsed()) / (iteration_ - device_->skip_iteration());
 
         // ss << forward_time << " , " << backward_time << std::endl;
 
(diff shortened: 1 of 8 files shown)


chrishkchris:
Results:
```
root@71ac539cda77:~/dcsysh/singa/examples/cnn# python3 benchmark.py --verbosity 1
100%|##################################################################################################################################################| 100/100 [00:15<00:00,  6.42it/s]
Throughput = 205.34910581511633 per second
TotalTime=15.583218574523926
Total=0.15583218574523927

Time Profiling:
Forward Propagation Time : 0.0456643 sec
Backward Propagation Time : 0.104244 sec
```

nudles:
@XJDKC pls help review.

XJDKC:
OK

src/core/scheduler/scheduler.cc
```diff
@@ -325,9 +337,10 @@ void Graph::RunGraph() {
     }
   }
 
+  EvaluateTimeElapsed();
+
   // increment iteration counter
   step();
```
XJDKC, line 343: Remember to clear the elapsed time and the iteration when resetting the graph.

chrishkchris: Addressed

Added iteration_=0 in Graph::reset, the elapsed time will be cleared in Graph::rest together with node as well

src/core/scheduler/scheduler.cc
```diff
@@ -325,9 +337,10 @@ void Graph::RunGraph() {
     }
   }
 
+  EvaluateTimeElapsed();
```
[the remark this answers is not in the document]
chrishkchris: Addressed

In the revised version, we call the step function before EvaluateTimeElapsed

chrishkchris:
Addressed the comments, a test:
```
root@71ac539cda77:~/dcsysh/singa/examples/cnn# python3 benchmark.py --verbosity 1
100%|##################################################################################################################################################| 100/100 [00:15<00:00,  6.62it/s]
Throughput = 211.74177390107678 per second
TotalTime=15.11274766921997
Total=0.1511274766921997

Time Profiling:
Forward Propagation Time : 0.0442915 sec
Backward Propagation Time : 0.101003 sec

```

Issue #700, opened 2020-05-17
Expand the model zoo (example model set)


nudles:
SINGA has multiple example models at http://singa.apache.org/docs/examples/
Some are implemented from scratch and some are converted from ONNX, which has a bigger model zoo  https://github.com/onnx/models.
The task is to convert more onnx models and implement some popular (and interesting) models that are not in onnx model zoo.
Here are some reference model zoos https://modelzoo.co/, https://gluon-nlp.mxnet.io/model_zoo/index.html

nudles:
Examples (To be updated):
MobileNetV2
ShuffleNetV2
YOLO
Mask RCNN
Faster RCNN
InceptionNetV3

Shashankwer:
Will Start working with MobileNetV2, InceptionNetV3

Alvinnyk:
I will work on Mask RCNN

joddiy:
Hi, guys, @Shashankwer , @Alvinnyk , thanks for your help. 

Before you start, please check this Jira, https://issues.apache.org/jira/projects/SINGA/issues/SINGA-509?filter=allopenissues

It lists the models which require support and for some of the models, SINGA may lack the necessary operators now. 

So I hope you:

1. check the Jira carefully, to see which model we can do now(I guess is: super_resolution, ResNet101_DUC_HDC, ShuffleNet_V2, ShuffleNet_V1, inception_v2, squeezenet).
2. find these models in ONNX model zoo, download them and open them by a tool called [`Netron`](https://github.com/lutzroeder/netron), try to understand the model and check each part refer the [IR doc](https://github.com/onnx/onnx/blob/master/docs/IR.md) of ONNX.
3. try to read the code of `SingaRep` and `SingaBackend` in `python/singa/soonx.py` to understand how we parse and construct the SINGA model.
4. try to read the code `examples/onnx`, to make out how we load and do the inference for a ONNX model.
5. then you can try to support a model(in first point) by yourself.

agnesnatasya:
I will try to work on the ShuffleNet

agnesnatasya:
Hi, may I ask how do we get the Amazon AWS link to download the model itself? For example, for ResNet, the link is https://s3.amazonaws.com/onnx-model-zoo/resnet/resnet18v1/resnet18v1.tar.gz. How do we get the link for the others? Because if I copy the link from ONNX's github, the link to download is Github's link, not Amazon AWS link. Thank you!

Shashankwer:
[...]

PR #711, opened 2020-05-27
Implement Squeezenet using Squeezenet1.1

agnesnatasya:
Implemented Squeezenet model using Squeezenet1.1 model in ONNX model zoo. 
As compared to the inference here https://github.com/onnx/models/blob/master/vision/classification/imagenet_inference.ipynb , which resizes the image to 256 and crop the 224 in the center, I resized the image to 224 and take all these 224 pixels as an input to the model. It gave different result when I tried these 2 approaches, but the second one gives more similar result. Thus, I choose the second one. Is there any significant impact of this difference?


nudles:
@joddiy  pls help review.

joddiy:
> @joddiy pls help review.

sure, I'll review it soon. I'm working on the `test_operations.py`. I guess I will review this PR today later.

@agnesnatasya, Thanks for your commit, if you're free, you can continue to the next model. Please leave a comment about the model you chose in the #700 .

joddiy:
@agnesnatasya Thanks for your help! I've reviewed the code, it's ready to merge into the master. @nudles 

And the result seems fine:

image

agnesnatasya:
Thank you @joddiy for the review, will make the change and will push again

agnesnatasya:
@joddiy The guide stated that "After the configuration, linting should be automatically applied when editing source code file". There is no warning in the "PROBLEMS" tab, but there are some warnings when I run pylint from terminal. However, other files like vgg16.py, also have these warnings when I run pylint on it. 

Have reformatted with yapf.

nudles:
@agnesnatasya @joddiy thanks!

PR #712, opened 2020-05-27
Implemented superresolution model

agnesnatasya:
Implemented the super-resolution model. It generated the picture and I think it is correct. However, it does not pass this 
```
# verify the test
    from utils import load_dataset
    inputs, ref_outputs = load_dataset(os.path.join('/tmp', 'super_resolution', 'test_data_set_0'))
    x_batch = tensor.Tensor(device=dev, data=inputs[0])
    outputs = model.forward(x_batch)
    for ref_o, o in zip(ref_outputs, outputs):
        np.testing.assert_almost_equal(ref_o, tensor.to_numpy(o), 4)
```
It fails on the second line, giving
`ValueError: cannot reshape array of size 451584 into shape (224,224)`

I assume that this test does not hold for this test data set as it has a different shape. Please correct me if I am wrong.


joddiy:
> Implemented the super-resolution model. It generated the picture and I think it is correct. However, it does not pass this
> 
> ```
> # verify the test
>     from utils import load_dataset
>     inputs, ref_outputs = load_dataset(os.path.join('/tmp', 'super_resolution', 'test_data_set_0'))
>     x_batch = tensor.Tensor(device=dev, data=inputs[0])
>     outputs = model.forward(x_batch)
>     for ref_o, o in zip(ref_outputs, outputs):
>         np.testing.assert_almost_equal(ref_o, tensor.to_numpy(o), 4)
> ```
> 
> It fails on the second line, giving
> `ValueError: cannot reshape array of size 451584 into shape (224,224)`
> 
> I assume that this test does not hold for this test data set as it has a different shape. Please correct me if I am wrong.

Hi, @agnesnatasya , which line reports this error? I guess the √451584 is 672, it must be the output shape of the model, but the 224 is the input shape.

By the way, why don't you use the Image library, but use the resizeimage? I guess the Image maybe better.

agnesnatasya:
@joddiy the `inputs, ref_outputs = load_dataset(os.path.join('/tmp', 'super_resolution', 'test_data_set_0'))` gives the error, after some debugging, I think it is because the output_*.pb has the dims specified as 224 x 224 but it actually has 451584 size. Thus, it cannot resize itself. Could there be an error with the output_.pb file itself?

joddiy:
> @joddiy the `inputs, ref_outputs = load_dataset(os.path.join('/tmp', 'super_resolution', 'test_data_set_0'))` gives the error, after some debugging, I think it is because the output_*.pb has the dims specified as 224 x 224 but it actually has 451584 size. Thus, it cannot resize itself. Could there be an error with the output_.pb file itself?

Yes, you are right, I've checked the code, and it seems the error displays at this line:
https://github.com/onnx/onnx/blob/master/onnx/numpy_helper.py#L55

The output should be 672*672, but it gives a shape of 224*224, which incurs the error.

It'd be fine you skip this verification, you can just use an image to show a demo inference.

By the way, since you found this problem of their model, I guess you can submit an Issue at ONNX model zoo's repo.

agnesnatasya:
@joddiy Thank you for checking! Okay, I will submit an Issue at ONNX's repo

chrishkchris:
@joddiy shall we merge this PR first and update the code according to the new onnx API

PR #713, opened 2020-05-29
impl new opt api

dcslin:
impl new opt api, decay scheduler constant class, optimzer class, sgd momentum, dampening, added unit test, backward compatibility tested


dcslin:
opened as separate PR for ease of review

dcslin:
checking linting

dcslin:
checking test cases errors

dcslin:
As highlighted by rulin, currently test case nesterov on gpu could not pass. Other tests are passing.
But in new API and backward compatibility perspective, it is ready for merged.
I am looking into the nesterov gpu case for now.

PR #722, opened 2020-06-05
cudnn lstm

dcslin:
implemented cudnn lstm, revised api compared to previous version, lstm layer take 1 tensor as input. added lstm model testing, as api is
different from previous rnn impl


dcslin:
1. revised api now take 1 tensor for all sequences.
2. added option: return_sequences, batch_first, bidirectional, num layers, dropout

a sample model training can be found at: https://github.com/apache/singa/blob/ad65d62325456af32a0ae27eac68ddd7ef43cd00/test/python/test_model.py#L122

a problem is that not so sure why this model could not overfit the toy dataset

dcslin:
masking is not yet implemented.

dcslin:
todo: for varying sequence length, in c++ need to dynamically update the cudnn tensor descriptor, xDesc and etc..

dcslin:
hi @nudles, sorry for the delay, there are still issues regarding the demo model, updates are:
0. added gensim as word2vec converter. it is not cleared how the pooling part of the model design is done in the paper. But from the reference model, it contracts the lstm output tensor by mean on sequence axis, which could be done by `autograd.reduce_mean()`
1. as loss function `L = max{0, M − cosine(q, a+) + cosine(q, a−)}` required two forward passes, then one backward pass, which is not supported by singa. Tried to concate the a+ and a- into {bs2, seq, embed} tensor and make model accept input like `(q, a+, a-)`. then in testing phase it is confusing because there is no label for answer.
2. tried to implemented a simplified version that subsituting loss function `L = max{0, M − cosine(q, a+) + cosine(q, a−)}`, with mseloss, then the model could be  trained with date in the format of `<q, a+, 1>, <q, a-, 0>`. but there is convergence problem.
3. advised by @joddiy , we could train the model with data format: one batch has two samples `<q,a+>` and `<q,a->` ordered alternatively, then we modified the loss function compute the loss for every batch of 2 samples(batch_index 0: `pos_sim`, batch_index 1:`neg_sim`), still checking.

nudles:
> 
> 
> hi @nudles, sorry for the delay, there are still issues regarding the demo model, updates are:
> 0. added gensim as word2vec converter. it is not cleared how the pooling part of the model design is done in the paper. But from the reference model, it contracts the lstm output tensor by mean on sequence axis, which could be done by `autograd.reduce_mean()`
> 
>     1. as loss function `L = max{0, M − cosine(q, a+) + cosine(q, a−)}` required two forward passes, then one backward pass, which is not supported by singa. Tried to concate the a+ and a- into {bs2, seq, embed} tensor and make model accept input like `(q, a+, a-)`. then in testing phase it is confusing because there is no label for answer.

during test, we only need to compute cosine(q, a) . there is no cosine(q, a+) or cosine(q, a-).
> 
>     2. tried to implemented a simplified version that subsituting loss function `L = max{0, M − cosine(q, a+) + cosine(q, a−)}`, with mseloss, then the model could be  trained with date in the format of `<q, a+, 1>, <q, a-, 0>`. but there is convergence problem.
> 
>     3. advised by @joddiy , we could train the model with data format: one batch has two samples `<q,a+>` and `<q,a->` ordered alternatively, then we modified the loss function compute the loss for every batch of 2 samples(batch_index 0: `pos_sim`, batch_index 1:`neg_sim`), still checking.

dcslin:
the last commit is for update purpose: **current issue**: could not converge, and performance is poor. Currently model is implemented according to bisltm, cos sim, customized loss structure. Internally, each batch, a- and a+ are encoded by lstm and split to calculate the similarity.

dcslin:
currently checking the model implementation and data transformation to figure out why the model does not converge.

dcslin:
tried to trouble shoot by changing the data transformation and modeling. Latest updated added support on varying batch size and seq length and make cudnn lstm more easy to trouble shoot.

python/singa/autograd.py
```diff
@@ -1190,6 +1190,41 @@ def cross_entropy(y, t):
...
 
 
+class QALSTMLoss(Operator):
+
+    def __init__(self, M=0.2):
+        super(QALSTMLoss, self).__init__()
+        self.M = M
+
+    def forward(self, pos, neg):
+        # L = max{0, M - cosine(q, a+) + cosine(q, a-)}
```
nudles, line 1200: @XJDKC  pls help check if the codes are compatible for training with graph enabled.

examples/qabot/model.py
```diff
@@ -0,0 +1,79 @@
...
+
+class MLP(model.Model):
+    def __init__(self):
+        super(MLP, self).__init__()
+        self.linear1 = layer.Linear(500)
+        self.relu = layer.ReLU()
+        self.linear2 = layer.Linear(2)
+
+    def forward(self, q, a):
+        q=autograd.reshape(q, (q.shape[0], -1))
```
XJDKC, line 67: Can use the layers in layer.py.(Reshape, Cat and SoftmaxCrossEntropy)

examples/qabot/model.py
```diff
@@ -0,0 +1,79 @@
...
+                                   num_layers=num_layers,
+                                   bidirectional=True,
+                                   return_sequences=False,
+                                   rnn_mode=rnn_mode,
+                                   batch_first=batch_first)
+
+    def forward(self, q, a_batch):
+        q = self.lstm_q(q) # BS, Hidden*2
+        a_batch = self.lstm_a(a_batch) # {2, hidden*2}
+        bs_a = int(a_batch.shape[0]/2) # cut concated a-a+ to half and half
```
XJDKC, line 45: Does bs_a keep changing during training?

examples/qabot/train.py
```diff
@@ -0,0 +1,162 @@
...
+    # m = MLP()
+    m.optimizer = opt.SGD()
+
+    tq = tensor.Tensor((1, q_seq_length, embed_size), dev, tensor.float32)
+    ta = tensor.Tensor((2, a_seq_length, embed_size), dev, tensor.float32)
+
+    tq.set_value(0.0)
+    ta.set_value(0.0)
+
+    m.compile([tq, ta], is_train=True, use_graph=False, sequential=False)
```
XJDKC, line 150: If the graph is enabled, can we train the model correctly?

src/model/operation/rnn.cc
```diff
@@ -0,0 +1,407 @@
...
+           x.device());
+  Tensor hy(Shape{h.num_layers * (h.bidirectional ? 2 : 1), h.batch_size,
+                  h.hidden_size},
+            x.device());
+  Tensor cy(Shape{h.num_layers * (h.bidirectional ? 2 : 1), h.batch_size,
+                  h.hidden_size},
+            x.device());
+
+  y.device()->Exec(
+      [&y, &hy, &cy, &x, &hx, &cx, &W, &h](Context *ctx) {
```
XJDKC, line 216: Can't submit the operator like this. All tensors that are not parameters should be passed by value(These tensors will be released, there will some segmentation faults if they are not passed by value). Make sure the correctness of read_blocks and write_blocks.
Can refer to [convolution.cc](https://github.com/apache/singa/blob/master/src/model/operation/convolution.cc#L590).

src/model/operation/rnn.cc
```diff
@@ -0,0 +1,407 @@
...
+                 h.hidden_size * (h.bidirectional ? 2 : 1)},
+           x.device());
+  Tensor hy(Shape{h.num_layers * (h.bidirectional ? 2 : 1), h.batch_size,
+                  h.hidden_size},
+            x.device());
+  Tensor cy(Shape{h.num_layers * (h.bidirectional ? 2 : 1), h.batch_size,
+                  h.hidden_size},
+            x.device());
+  y.device()->Exec(
+      [&y, &hy, &cy, &x, &hx, &cx, &W, &h](Context *ctx) {
```
XJDKC, line 271: Same as the problem mentioned above.

src/model/operation/rnn.cc
```diff
@@ -0,0 +1,407 @@
...
+       W.block()},
+      {dx.block(), dhx.block(), dcx.block()});
+  return {dx, dhx, dcx};
+}
+
+Tensor GpuRNNBackwardW(const Tensor &x, const Tensor &hx, const Tensor &y,
+                       CudnnRNNHandle &h) {
+  Tensor dW(Shape{h.weights_size}, x.device());
+  dW.device()->Exec(
+      [&dW, &x, &hx, &y, &h](Context *ctx) {
```
XJDKC, line 380: Same as the problem mentioned above. And it seems that we have not yet implemented the CPU version.

python/singa/autograd.py
```diff
@@ -1190,6 +1190,41 @@ def cross_entropy(y, t):
...
+        zero = singa.Tensor(list(pos.shape()), pos.device())
+        zero.SetFloatValue(0.0)
+        val = singa.AddFloat(singa.__sub__(neg, pos), self.M)
+        gt_zero = singa.__gt__(val, zero)
+        self.inputs = (gt_zero, ) # (BS,)
+        all_loss = singa.__mul__(gt_zero, val)
+        loss = singa.SumAll(all_loss)
+        loss /= (pos.shape()[0])
+        # assert loss.shape(0) == 1
+        return loss
```
XJDKC, line 1210: 1. What's the type of neg and pos?
2. Just use neg-pos.
3. What's the type of gt_zero? If its type is boolean, it may not be buffered in the graph.

src/model/operation/rnn.cc
nudles: CPU version is not urgent. We will use GPU.

python/singa/autograd.py
dcslin: 1. neg pos are tensors of shape (bs,)
2. do you mean replace singa.__sub__ with "-"?
3. gt_zero is tensor of (bs,) value is float

src/model/operation/rnn.cc
dcslin: ok let me check again

examples/qabot/train.py
dcslin: setting use_graph to True gives:
```
  File "/root/singa/examples/qabot/train.py", line 184, in <module>
    training(m, train_data, max_epoch)
  File "/root/singa/examples/qabot/train.py", line 126, in training
    score, l = m(tq, ta)
  File "build/python/singa/model.py", line 210, in __call__
    return self.train_one_batch(*input, **kwargs)
  File "build/python/singa/model.py", line 63, in wrapper
    self._results = func(self, *args, **kwargs)
  File "/root/singa/examples/qabot/model.py", line 65, in train_one_batch
    self.optimizer.backward_and_update(loss)
  File "/usr/local/lib/python3.6/dist-packages/deprecated/classic.py", line 281, in wrapper_function
    return wrapped_(*args_, **kwargs_)
  File "build/python/singa/opt.py", line 320, in backward_and_update
    super(SGD, self).__call__(loss)
  File "build/python/singa/opt.py", line 103, in __call__
    self.call(loss)
  File "build/python/singa/opt.py", line 110, in call
    self.apply(p.name, p, g)
  File "build/python/singa/opt.py", line 270, in apply
    singa.Axpy(minus_lr.data, param_grad.data, param_value.data)
TypeError: in method 'Axpy', argument 1 of type 'float'
```

examples/qabot/model.py
dcslin: most of the time bs_a does not change, some times changes

examples/qabot/model.py
dcslin: ok

PR #727, opened 2020-06-09
cudnn lstm test case update

dcslin:


chrishkchris:
Since it is just a simple update of test case that won't affect any operation, I merge it for testing first

PR #734, opened 2020-06-14
add pad operator

joddiy:


python/singa/autograd.py
```diff
@@ -4698,6 +4698,118 @@ def cossim(a, b):
...
 
+class Pad(Operator):
+    """
+    Init a pad operator
+    """
+
+    def __init__(self, mode, pads, constant=0.):
+        """
+        Args:
+            mode (string): Supported modes: `constant`(default), `reflect`, `edge`.
```
nudles, line 4709: what does reflect  and edge mean?

python/singa/autograd.py
```diff
@@ -4698,6 +4698,118 @@ def cossim(a, b):
...
+            pads (list[int]): Tensor of integers indicating the number of padding elements
```
nudles, line 4710: pads is not a tensor.. it is a list of int.

python/singa/autograd.py
joddiy: fixed, please check the example usage

PR #736, opened 2020-06-15
Add expand operator

joddiy:


python/singa/autograd.py
```diff
@@ -4698,6 +4698,103 @@ def cossim(a, b):
...
+                        dxs.append(tmp_tensor)
+                    dxs = singa.VecTensor(dxs)
+                    dy = singa.ConcatOn(dxs, axis)
+        dy = singa.Reshape(dy, self.x_shape)
+        return dy
+
+
+def expand(x, shape):
+    """
+    Produces a cos similarity operator
```
nudles, line 4787: expand and cos similarity are related??

python/singa/autograd.py
```diff
@@ -4698,6 +4698,103 @@ def cossim(a, b):
     return CosSim()(a, b)[0]
 
 
+class Expand(Operator):
+    """
+    Init a expand operator
```
nudles, line 4703: You can explain what this operator does?
Does it reshape the tensor with more dimensions?

nudles: what's the difference to reshape?

python/singa/autograd.py
joddiy: just a typo, fixed

python/singa/autograd.py
joddiy: I've added an example usage at the comment.

The `expand` duplicates the data along with the axis. It looks like `upsample` but still has some differences.
>
    Example usage::
    data = [[1.], [2.], [3.]]

    # dim_changed
    shape = [2, 1, 6]
    output = [[[1., 1., 1., 1., 1., 1.], 
               [2., 2., 2., 2., 2., 2.],
               [3., 3., 3., 3., 3., 3.]],
              [[1., 1., 1., 1., 1., 1.],
               [2., 2., 2., 2., 2., 2.],
               [3., 3., 3., 3., 3., 3.]]]

    # dim_unchanged
    shape = [3, 4]
    output = [[1., 1., 1., 1.],
              [2., 2., 2., 2.],
              [3., 3., 3., 3.]]
>

For upsample
>
    Example usage::
    data = [[[[1, 2],
              [3, 4],]]]

    # nearest
    scales = [1.0, 1.0, 2.0, 3.0]
    output = [[[[1, 1, 1, 2, 2, 2],
                [1, 1, 1, 2, 2, 2],
                [3, 3, 3, 4, 4, 4],
                [3, 3, 3, 4, 4, 4],]]]
>

PR #737, opened 2020-06-15
Add upsample operator

joddiy:


python/singa/autograd.py
```diff
@@ -4698,6 +4698,82 @@ def cossim(a, b):
     return CosSim()(a, b)[0]
 
 
+class UpSample(Operator):
+    """
+    Init a UpSample operator
```
nudles, line 4703: why use "Init" here?

python/singa/autograd.py
```diff
@@ -4698,6 +4698,82 @@ def cossim(a, b):
...
+    def __init__(self, mode, scales):
+        """
+        Args:
+            scales (list[int]): The scale array along each dimension. It takes 
+                value greater than or equal to 1. 
+        """
+        super(UpSample, self).__init__()
+        self.scales = scales
+        self.mode = mode
+        if self.mode != "nearest":
```
nudles, line 4715: better use self.mode.lower() != "nearest"

python/singa/autograd.py
```diff
@@ -4698,6 +4698,82 @@ def cossim(a, b):
...
+        for axis, s in zip(range(len(self.scales)), self.scales):
+            s = int(s)
+            if s == 1:
+                continue
+            x = x.Repeat([s, ], axis)
+        return x
+
+    def backward(self, dy):
+        """
+        backward of UpSample
```
nudles, line 4740: this kind of docstring is trivial and thus is not necessary.
we do not need docstring for every function/class/variable if the name is self-explainable

python/singa/autograd.py
joddiy: fixed

python/singa/autograd.py
joddiy: fixed

python/singa/autograd.py
joddiy: fixed

Issue #698, opened 2020-05-16
Implement popular initialization methods


nudles:
Add each initialization method as a function or class in https://github.com/apache/singa/blob/master/python/singa/initializer.py. e.g.,
```python
class InitializationBase(object):
      def __init__(self, ):
          pass
      def call(self, x):
          pass
      def __call__(self, x):
          self.call(x)

class Uniform(InitializationBase):
       def __init__(self, low=-1, high=1):
               self.low = low
               self.high = high
      def call(self, x):
              x.uniform(self.low, self.high)

def uniform(x, low=-1, high=1):
       x.uniform(low, high)
```

nudles:
resolved in PR#739

PR #739, opened 2020-06-17
Update the initializer module to add more initialization methods.

nudles:
This is a solution for Issue #698.
Add notice to some deprecated modules and test scripts.


chrishkchris:
I helped to double check the test case in my environment 

```
root@d05828f767ee:~/dcsysh/singa2/test/python# python3 test_initializer.py
....
Ran 4 tests in 0.027s

OK
```

Could I merge this PR?

PR #742, opened 2020-06-18
fix onnx backend cuda cases

joddiy:
Fix onnx backend cuda cases when there is not a gpu device.


joddiy:
@nudles urgent to merge

PR #753, opened 2020-06-29
fix test operation Tensor class non hashable erorr

dcslin:


python/singa/autograd.py
```diff
@@ -117,7 +117,11 @@ def gradients(y, dy=None):
     """
     grads = {}  # mapping: x->dx if x.stores_grad
     for p, dp in backward(y, dy):
-        grads[p] = dp
+        # TODO: this fn is only helper for test case for now.
+        #   1. could implement __hash__ or
```
nudles, line 121: does it affect other places e.g., optimizer, which will take the param and grad for updating.

dcslin: This `gradients()` is currently only used in 2 of the operations test cases, so it does not affect other places.
`Opt.py` does backward itself, and not using this function.
This `gradients()` is more like "do backward and do NOT update and return gradients". I did not directly move it into test scripts because in some cases, user might want to get gradiants explicitly. Like pytorch `x.grad`.

nudles: ok. then pls update the docstring for the returned data.

PR #754, opened 2020-06-30
Fix SINGA-gpu conda build and update numpy version

chrishkchris:
1. When I was building conda package of singa-gpu, I encountered the following error

building

One available solution is to call `roundf` in CUDA MATH API instead of `std::round `
https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html#group__CUDA__MATH__SINGLE_1ga1c1521079e51b4f54771b16a7f8aeea

With this modification, the round test case can pass:
```
root@e7f452d7c67a:~/dcsysh/singa/test/python# python3 test_api.py
.................................
Ran 33 tests in 1.409s

OK
```
By the way, in the docker version of singa, it does not need this change, maybe it is the difference of GCC version in the two environments.

2. This PR updates the numpy to 1.18.5

3. Seems the `cudaDeviceReset()` in destructor returns "driver shutting down" code, we don't need the `CUDA_CHECK`.

```
root@e7f452d7c67a:~/dcsysh/singa/test/python# python3 test_api.py
.................................
Ran 33 tests in 1.383s

OK
WARNING: Logging before InitGoogleLogging() is written to STDERR
F0630 12:40:39.892601  3077 cuda_gpu.cc:49] Check failed: error == cudaSuccess (29 vs. 0)  driver shutting down
*** Check failure stack trace: ***
Aborted (core dumped)
```
After changed, it works well
```
root@e7f452d7c67a:~/dcsysh/singa/test/python# python3 test_api.py
.................................
Ran 33 tests in 1.564s

OK
root@e7f452d7c67a:~/dcsysh/singa/test/python#
```

### DIFF
--- a/tool/conda/singa/meta.yaml
+++ b/tool/conda/singa/meta.yaml
@@ -50,7 +50,7 @@ requirements:
     - protobuf 3.10.0         # [osx]
     - protobuf 3.9.2          # [linux]
     - glog 0.3.5
-    - numpy 1.18.5
+    - numpy >=1.16,<2.0
     - pytest
     - deprecated 1.2.7
     - cudnn {{ cudnn }}       # ['cudnn' in str(build_str)]
--- a/tool/conda/singa/meta.yaml
+++ b/tool/conda/singa/meta.yaml
@@ -50,7 +50,7 @@ requirements:
     - protobuf 3.10.0         # [osx]
     - protobuf 3.9.2          # [linux]
     - glog 0.3.5
-    - numpy 1.16.5
+    - numpy 1.18.5
     - pytest
     - deprecated 1.2.7
     - cudnn {{ cudnn }}       # ['cudnn' in str(build_str)]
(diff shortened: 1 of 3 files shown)


tool/conda/singa/meta.yaml
```diff
@@ -50,7 +50,7 @@ requirements:
     - protobuf 3.10.0         # [osx]
     - protobuf 3.9.2          # [linux]
     - glog 0.3.5
-    - numpy 1.16.5
```
nudles, line 53: I think we can use a range like >=1.16,<2.0

chrishkchris: Yes, updated the code. The travis CI build the conda package sucessfully 
https://travis-ci.org/github/apache/singa/jobs/703751043#L2518

chrishkchris:
Yes, updated the code. The travis CI build the conda package sucessfully
https://travis-ci.org/github/apache/singa/jobs/703751043#L2518

Ready for merge

PR #755, opened 2020-06-30
Add onnx gpt2 model

joddiy:
- The output is correct now,
- But cannot generate more words, 
- Mostly it is an issue of onnx model, not sonnx, still checking


joddiy:
Yes, it definitely has some problem in onnx model, I've submitted a PR to onnx.

https://github.com/onnx/models/issues/326

joddiy:
works fine now, ready to merge @nudles 
image

PR #758, opened 2020-07-02
update roundeven backend algorithm

dcslin:


dcslin:
HI @joddiy , this is updated according to the round even requirement, could you please help to review?

chrishkchris:
error message at math_kernel.cu line 143:

```
[ 14%] Building NVCC (Device) object src/CMakeFiles/cuda_compile_1.dir/core/tensor/cuda_compile_1_generated_math_kernel.cu.o
/root/miniconda/conda-bld/singa_1593705642277/work/src/core/tensor/math_kernel.cu(143): error: explicit type is missing ("int" assu                                                               med)

1 error detected in the compilation of "/tmp/tmpxft_00000574_00000000-6_math_kernel.cpp1.ii".
CMake Error at cuda_compile_1_generated_math_kernel.cu.o.cmake:280 (message):
  Error generating file
  /root/miniconda/conda-bld/singa_1593705642277/work/build/src/CMakeFiles/cuda_compile_1.dir/core/tensor/./cuda_compile_1_generated                                                               _math_kernel.cu.o
```

Maybe I try later version of GCC first

chrishkchris:
> error message at math_kernel.cu line 143:
> 
> ```
> [ 14%] Building NVCC (Device) object src/CMakeFiles/cuda_compile_1.dir/core/tensor/cuda_compile_1_generated_math_kernel.cu.o
> /root/miniconda/conda-bld/singa_1593705642277/work/src/core/tensor/math_kernel.cu(143): error: explicit type is missing ("int" assu                                                               med)
> 
> 1 error detected in the compilation of "/tmp/tmpxft_00000574_00000000-6_math_kernel.cpp1.ii".
> CMake Error at cuda_compile_1_generated_math_kernel.cu.o.cmake:280 (message):
>   Error generating file
>   /root/miniconda/conda-bld/singa_1593705642277/work/build/src/CMakeFiles/cuda_compile_1.dir/core/tensor/./cuda_compile_1_generated                                                               _math_kernel.cu.o
> ```
> 
> it needs explicit type instead of auto?
> 
> Maybe I try later version of GCC first

Resolved after upgrading C and CXX complier versions in https://github.com/apache/singa/pull/759
-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0

src/core/tensor/math_kernel.cu
```diff
@@ -140,7 +140,12 @@ __global__ void KernelRound(const size_t n, const float *in, float *out) {
 __global__ void KernelRoundE(const size_t n, const float *in, float *out) {
   for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < n;
        i += blockDim.x * gridDim.x) {
-    out[i] = roundf(in[i]/2)*2;
+    auto doub = in[i]*2;
```
chrishkchris, line 143: This auto causes problem in C complier version 5.4, upgraded to 7.5.0 will then be okay. Resolved in #759

dcslin: Thank you Chris for the prompt response! I could also replace the `auto` in this PR. Yes I think updating to newer gcc is a good idea, although the "new"(c++11) one is already widely used and not that new, it would be helpful to support running c++11 code.

chrishkchris: I guess there is no need to change the `auto`, `auto` seems to be a good way to prevent implicit type casting (sometime developer us could be careless). The only problem of `auto` is just the code readability for the developer to know the exact datatype.

In any case, I think need to keep the same version of complier between the docker version and conda version (as in #759), because a few day ago we had the similar complier error because of the version different.

chrishkchris:
@dcslin The test case looks good to me
https://blogs.sas.com/content/iml/2019/11/11/round-to-even.html
As the example given in the link above, `rounde` rounds 0.5 to 0, and 2.5 to 2
The test case passed so it should be correct.
```
root@71ac539cda77:~/dcsysh/singa/test/python# python3 test_api.py
.................................
Ran 33 tests in 0.758s

OK
```

PR #761, opened 2020-07-08
C++ build with Github Actions

moazreyad:
This is the first build with Github Actions which can be used to replace Travis CI.

### DIFF
--- a/.github/workflows/cpp.yaml
+++ b/.github/workflows/cpp.yaml
@@ -55,7 +55,7 @@ jobs:
       - name: get-oneDNN
         run: wget https://github.com/oneapi-src/oneDNN/releases/download/v1.1/dnnl_lnx_1.1.0_cpu_gomp.tgz -P /tmp/ && tar zxf /tmp/dnnl_lnx_1.1.0_cpu_gomp.tgz
       - name: install-build-dependencies
-        run: sudo apt-get install -y libgoogle-glog-dev libprotobuf-dev protobuf-compiler libncurses-dev libopenblas-dev gfortran libblas-dev liblapack-dev libatlas-base-dev swig libcurl3-dev cmake dh-autoreconf && wget https://github.com/oneapi-src/oneDNN/releases/download/v1.1/dnnl_lnx_1.1.0_cpu_gomp.tgz -P /tmp/ \ && tar zxf /tmp/dnnl_lnx_1.1.0_cpu_gomp.tgz -C /root
+        run: sudo apt-get install -y libgoogle-glog-dev libprotobuf-dev protobuf-compiler libncurses-dev libopenblas-dev gfortran libblas-dev liblapack-dev libatlas-base-dev swig libcurl3-dev cmake dh-autoreconf
       - name: configure
         run: mkdir build && cd build && cmake -DUSE_PYTHON=NO -DENABLE_TEST=YES -DUSE_DNNL=YES ..
         env:
--- a/.github/workflows/cpp.yaml
+++ b/.github/workflows/cpp.yaml
@@ -1,0 +1,47 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# This is a basic workflow to help you get started with Actions
+
+name: C++
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches:
+      - master
+      - dev
+  pull_request:
+    branches:
+      - master
+      - dev
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  build-ubuntu-cpp:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: install-build-dependencies
+        run: sudo apt-get install -y libgoogle-glog-dev libprotobuf-dev protobuf-compiler libncurses-dev libopenblas-dev gfortran libblas-dev liblapack-dev libatlas-base-dev swig libcurl3-dev cmake dh-autoreconf  
+      - name: configure
+        run: mkdir build && cd build && cmake -DUSE_PYTHON=NO ..
+      - name: build
+        run: cd build && make
+       
--- a/.github/workflows/cpp.yaml
+++ b/.github/workflows/cpp.yaml
@@ -53,7 +53,7 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - name: get-oneDNN
-        run: wget https://github.com/oneapi-src/oneDNN/releases/download/v1.1/dnnl_lnx_1.1.0_cpu_gomp.tgz -P /tmp/ && tar zxf /tmp/dnnl_lnx_1.1.0_cpu_gomp.tgz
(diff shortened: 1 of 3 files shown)


moazreyad:
The team may discuss and decide when should we run the C++ build action. In this example, I assume we run it on push and pull requests on Master and Dev branches. But of course this can be changed if the team prefers other options.

Currently it tests the build on latest ubuntu, we can also build on previous versions of ubuntu if needed. but it can be too much for each pull request. what do you think? may be the complete test against all the versions of ubuntu can be enabled on the Master branch only, while a quick test on the latest ubuntu can be done with each pull request on dev? or on all branches except Master?

We also need to run the C++ and Python tests. However both of them fail now and if I add them here, they fail the Github check. The C++ test error can be found [here](https://github.com/moazreyad/singa/runs/855402800?check_suite_focus=true#step:6:98). Can anyone fix this error and make sure that the C++ test passes before we add it to this workflow of the Github Action? The problem happens when we compile SINGA with ENABLE_TEST=ON, and  run the unit tests by ./bin/test_singa

nudles:
> 
> 
> The team may discuss and decide when should we run the C++ build action. In this example, I assume we run it on push and pull requests on Master and Dev branches. But of course this can be changed if the team prefers other options.
> 
> Currently it tests the build on latest ubuntu, we can also build on previous versions of ubuntu if needed. but it can be too much for each pull request. what do you think? may be the complete test against all the versions of ubuntu can be enabled on the Master branch only, while a quick test on the latest ubuntu can be done with each pull request on dev? or on all branches except Master?

+1.

> 
> We also need to run the C++ and Python tests. However both of them fail now and if I add them here, they fail the Github check. The C++ test error can be found [here](https://github.com/moazreyad/singa/runs/855402800?check_suite_focus=true#step:6:98). Can anyone fix this error and make sure that the C++ test passes before we add it to this workflow of the Github Action? The problem happens when we compile SINGA with ENABLE_TEST=ON, and run the unit tests by ./bin/test_singa

nudles:
@chrishkchris could you help check the cpp test cases? thanks!

chrishkchris:
> @chrishkchris could you help check the cpp test cases? thanks!

Thanks, I will look at the problem in cpp test case and submit a PR to fix

.github/workflows/cpp.yaml
```diff
@@ -0,0 +1,47 @@
...
+jobs:
+  build-ubuntu-cpp:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: install-build-dependencies
+        run: sudo apt-get install -y libgoogle-glog-dev libprotobuf-dev protobuf-compiler libncurses-dev libopenblas-dev gfortran libblas-dev liblapack-dev libatlas-base-dev swig libcurl3-dev cmake dh-autoreconf  
+      - name: configure
+        run: mkdir build && cd build && cmake -DUSE_PYTHON=NO ..
```
chrishkchris, line 44: Since we are using DNNL as one of the C++ computational kernal (other example of comp kernals are BLAS etc.)

The cmake flag should include -DUSE_DNNL=ON

FYI:

1. This is the cmake command in conda version of SINGA
https://github.com/apache/singa/blob/dev/tool/conda/singa/build.sh#L43
```
cmake -DCMAKE_INSTALL_PREFIX=$PREFIX -DUSE_CUDA=$USE_CUDA \
	-DUSE_PYTHON3=ON -DUSE_DNNL=ON -DUSE_DIST=$USE_DIST -DCMAKE_OSX_SYSROOT=${CONDA_BUILD_SYSROOT} ..
```


2. This is the cmake command in docker version of SINGA
https://github.com/apache/singa/blob/dev/tool/docker/devel/ubuntu/cpu/Dockerfile#L67
`cmake -DENABLE_TEST=ON -DUSE_PYTHON3=ON -DUSE_DNNL=ON ..`

You may consider refer to both versions on how to include the DNNL library

Thanks!

.github/workflows/macOS.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.7' 
+      - name: install-build-dependencies  
+        run: brew install protobuf swig opencv glog lmdb numpy && pip3 install numpy
+      - name: configure        
+        run: mkdir build && cd build && cmake -DUSE_PYTHON3=YES ..
```
chrishkchris, line 36: the comment to the cmake flag is the same as above

.github/workflows/cpp.yaml
moazreyad: Thank you for the revision, here are some comments:

1. The MKL-DNN is now called [oneDNN](https://github.com/oneapi-src/oneDNN)

2. If the flag -DUSE_DNNL is always required to pass the C++ tests, then why it is an optional flag? It should be always ON by default since the system does not work without it. (i.e. the system does not pass the unit tests without it)

3. If the C++ unit tests can work without the -DUSE_DNNL flag except few unit tests, then I propose to add "ifdef USE_DNNL" around the unit tests that require DNNL:

```
#ifdef USE_DNNL

.. unit test code

#endif
```


So that these unit tests are not compiled if they don't support the given configuration to cmake.

What do you think?

chrishkchris: Thanks so much for your ideas. I will think about your solutions 2 and 3, and discuss with others.

I think the DNNL (oneDNN) is very important in the test. Is the Github Actions code able to provide the library for test? We are using DNNL version 1.1 as indicated in https://github.com/apache/singa/blob/dev/tool/docker/devel/ubuntu/cpu/Dockerfile#L52

chrishkchris:
@moazreyad 

This PR can fix the test case without DNNL (by adding back the comp kernal without DNNL using the else function)
https://github.com/apache/singa/pull/765
Meanwhile, I don't know which python test case are failed, you may provide the list of python failed test case

moazreyad:
> @moazreyad
> 
> This PR #765 can fix the test case without DNNL (by adding back the comp kernal without DNNL using the else function)
> 
Very good. If the PR #765 can be merged with dev, let's merge it. We can also merge this PR #761 which checks the build on latest ubuntu and macOS. Then after that I can create another PR for running the C++ unit tests on ubuntu and mac. Is this ok?

moazreyad:
> Meanwhile, I don't know which python test case are failed, you may provide the list of python failed test case

There are many pyhton skipped and failed tests in Travis CI build for [ubuntu](https://travis-ci.org/github/apache/singa/jobs/707332892#L4558) and [macOS](https://travis-ci.org/github/apache/singa/jobs/707332889#L5274):

```
FAILED (failures=5, errors=8, skipped=1067)

TEST CONV2D FORWARD

TEST CONV2D DATA BACKWARD

TEST CONV2D WEIGHT BACKWARD

TEST CONV2D DATA BACKWARD
```

chrishkchris:
> > Meanwhile, I don't know which python test case are failed, you may provide the list of python failed test case
> 
> There are many pyhton skipped and failed tests in Travis CI build for [ubuntu](https://travis-ci.org/github/apache/singa/jobs/707332892#L4558) and [macOS](https://travis-ci.org/github/apache/singa/jobs/707332889#L5274):
> 
> ```
> FAILED (failures=5, errors=8, skipped=1067)
> 
> TEST CONV2D FORWARD
> 
> TEST CONV2D DATA BACKWARD
> 
> TEST CONV2D WEIGHT BACKWARD
> 
> TEST CONV2D DATA BACKWARD
> ```

ok, thanks, currently I am fixing some python test cases first

moazreyad:
C++ tests passes successfully now on ubuntu in this PR. (see [here](https://github.com/apache/singa/pull/761/checks?check_run_id=869557107#step:6:448))

.github/workflows/cpp.yaml
moazreyad: Yes, any shell script for build can be used. I added the [oneDNN](https://github.com/apache/singa/pull/761/checks?check_run_id=871284806) check and it works fine.

chrishkchris:
Thanks, the two C++ test (one with DUSE_DNNL=ON, another with OFF) passes

chrishkchris:
@moazreyad I have some little discussion with them. Could we turn off the C++ test without oneDNN, because it will make the C++ test faster to be completed (no need to test both)

moazreyad:
I turned off C++ test without oneDNN and made the build-ubuntu-oneDNN available on all branches (not only master and dev). The team can change the settings later as needed. For example, may be to enable all the tests on master branch, and only few on the other branches.

moazreyad:
For the C++ test on macOS, the build does not work with -DENABLE_TEST=YES on macOS. (see [here](https://github.com/moazreyad/singa/runs/871231220?check_suite_focus=true))  due to a missing `#include <array>` in `test_tensor_math.cc` So we test only the build without testing on macOS in this PR. I will make another PR for macOS test fix.

PR #764, opened 2020-07-13
Qabot fixes

dcslin:


dcslin:
diff against previous qabot model:
1. auto download data
2. data cleaning
3. fix cos sim operator, add output value check numpy implementation for cos sim
4. cudnn rnn output value check, gradient check
5. qabot loss function output value check
6. added a util tensor operation tensor.random, for quick testing
7. set individual cudnn rnn weights by new api

problem to check: pooling 1D layer has some issue

PR #766, opened 2020-07-14
Fix python test cases

chrishkchris:
Continue to fix the remaining python test cases


dcslin:
yes this is looking ok to me.

chrishkchris:
ready for merge, thanks

PR #775, opened 2020-08-03
half float

dcslin:


dcslin:
working on testing python singa APIs.

dcslin:
refer to https://github.com/apache/singa/pull/779

PR #780, opened 2020-08-18
Adding dependabot support

moazreyad:
Dependabot is a native tool in github that automatically checks the dependencies of the project and creates pull requests for outdated or insecure dependencies. See how it works from [here](https://dependabot.com/#how-it-works).

Currently it will analyze only the dependencies defined in [pom.xml ](https://github.com/apache/singa/blob/master/java/pom.xml)in SINGA because other dependencies are hard coded in configuration or shell scripts which are not supported by this tool. In future commits, the dependencies will be moved to standard files such as python requirements.txt so that they become available to dependabot and similar tools.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+version: 2
+updates:
+  - package-ecosystem: "" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "daily"


nudles:
shall we move all CI functions from travis to github to avoid the duplicated testing?

moazreyad:
> shall we move all CI functions from travis to github to avoid the duplicated testing?

Yes, I will create another PR for the work in progress on replacing travis with github actions.

moazreyad:
Both java and python dependencies are enabled now. Dependabot will check python daily and java monthly. It will create a PR for upgrade insecure or outdated dependencies in SINGA like this [java PR](https://github.com/moazreyad/singa/pull/19) and this [python PR](https://github.com/moazreyad/singa/pull/21).

notes:
- Dependabot works only on the master branch. It will be enabled after merging with master.
- The proposed requirements.txt for python is not complete. Later we need to add all the requirements for all the configurations.

nudles:
let's consider this PR in the next version.
Now we support both conda and pip.
One potential issue is that protocol buffer (PB) is not backward compatible. 
If there is a new version of PB, we may have to update many dependent libs to remove the warnings from the dependent bot.

moazreyad:
> let's consider this PR in the next version.

Ok. Although we can at least enable the dependabot support and badge, even though we don't resolve the problems that dependabot report. Just like we enabled lgtm but we don't have to fix all its errors, and we enabled codecov but we don't have to test with 100% coverage. 

> One potential issue is that protocol buffer (PB) is not backward compatible.
> If there is a new version of PB, we may have to update many dependent libs to remove the warnings from the dependent bot.

We don't have to update immediately all the dependencies that dependabot found obsolete or insecure. At least we can have the results of the dependency analysis and the singa team can decide the work priorities. The PB team are working now on a release candidate for version 4 and planning to upgrade singa PB to at least version 3 can be useful.

> Now we support both conda and pip.

We don't support machine readable requirements that can be found by Github tools. The instructions to installing singa dependencies with pip is only human readable in the web site documentation, or in the conda scripts. We need to extract these dependencies into something like requirements.txt so that Github (and other tools) can find them.

PR #789, opened 2020-09-05
fixed imdb train script args

dcslin:


dcslin:
Hi @chrishkchris , this is regarding the imdb training script issue, could please help to review, options are working now:
return sequences:
```
root@1c6aaef3db53:~/singa# PYTHONPATH=build/python/ python3 examples/rnn/imdb_train.py -s
epoch 0 loss [0.6632729]; acc 0.593
epoch 1 loss [0.5689068]; acc 0.705
```
no return sequences:
```
root@1c6aaef3db53:~/singa# PYTHONPATH=build/python/ python3 examples/rnn/imdb_train.py -s
epoch 0 loss [0.6632729]; acc 0.593
epoch 1 loss [0.5689068]; acc 0.705
```

chrishkchris:
I remembered last time is this problem?

577d5bfda72d47956db1b3439705862
d32bb0f4bdef62a3f989e53e8b4c15d

dcslin:
> I remembered last time is this problem?
> 
> 577d5bfda72d47956db1b3439705862
> d32bb0f4bdef62a3f989e53e8b4c15d

Hi Chirs, I think the bug was because the `argpaser` configuration was wrong. And the `args.return_sequences` has string value of `"False"` which is evaluated to `True`(not `None`) in `if` statement. So the changes in this PR rectified the `argpaser` to make `args.return_sequences` has boolean value.

PR #790, opened 2020-09-06
Remove Travis CI

moazreyad:
This PR removes Travis CI since we are using Github Actions now. :octocat:

- We will not run the cpp and python lint commands:

tool/linting/py.sh
tool/linting/cpp.sh

Instead we can just use the output of Github's lgtm. Is this fine with everyone?

- We do not have a github workflow that publishes the conda packages. If this is required, we can add the [Conda Package Publish Action](https://github.com/marketplace/actions/conda-package-publish-action) at the end of the conda workflow. Assuming there is secrets.ANACONDA_PASSWORD and secrets.ANACONDA_USERNAME defined in the [github secrets](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets) of the user who runs this workflow.

- Currently the github workflows use the latest Ubuntu only. If it is required to support another older versions, we can add them to the workflow. Travis CI was checking Ubuntu 14.04, 16.04 and 18.04.


chrishkchris:
I have a question: the current CI use the current PR branch for CI test. 

In this case, if one PR fixes a bug that cause a failed CI test, some other PR still cannot pass the test unless rebased to dev branch

Would it be a better practice if we use the merged branch instead of the PR branch?

moazreyad:
> Would it be a better practice if we use the merged code instead of the PR branch for CI test?

It is possible to remove `pull_request:` from the trigger events in the [workflow yaml files](https://github.com/apache/singa/tree/dev/.github/workflows):

```
on:
  push:
  pull_request:
```

or we can keep it as a guide for the PR reviewer but not requiring all PR to pass all the tests. It will be up to the PR reviewer to decide if the failed tests should be fixed in this PR or not. I would propose to keep it because sometimes it is better to discover the failed tests in the PR before the merge. But all the team can vote on the best options for triggering the workflow. There are many [events for workflow trigger](https://docs.github.com/en/actions/reference/events-that-trigger-workflows) and may be some of them can be enabled on some specific workflows.

> Also another problem is that if team members A and B working on the same branch, their code can pass the CI test separately, but may not be compatible to each other (e.g. due to API change)

If the `push:` trigger is enabled, this problem will be discovered when the second team member pushes the code and the workflow is triggered.

chrishkchris:
> > Would it be a better practice if we use the merged code instead of the PR branch for CI test?
> 
> It is possible to remove `pull_request:` from the trigger events in the [workflow yaml files](https://github.com/apache/singa/tree/dev/.github/workflows):
> 
> ```
> on:
>   push:
>   pull_request:
> ```
> 
> or we can keep it as a guide for the PR reviewer but not requiring all PR to pass all the tests. It will be up to the PR reviewer to decide if the failed tests should be fixed in this PR or not. I would propose to keep it because sometimes it is better to discover the failed tests in the PR before the merge. But all the team can vote on the best options for triggering the workflow. There are many [events for workflow trigger](https://docs.github.com/en/actions/reference/events-that-trigger-workflows) and may be some of them can be enabled on some specific workflows.
> 
> > Also another problem is that if team members A and B working on the same branch, their code can pass the CI test separately, but may not be compatible to each other (e.g. due to API change)
> 
> If the is enabled, this problem will be discovered when the second team member pushes the code and the workflow is triggered.

Thanks a lot for the opinion. I also think that keeping all the test is good, as it helps us to safeguard the code base.

I suggest using the merged code (i.e. the code merging from PR branch and dev branch) for CI test instead of using PR branch code only for CI test, the flow I suggest is something like this:
Step 1: Merged_code = Merge (PR branch, base branch)
Step 2: Test_result = CI_Test(Merged_code)
Step 3: Make decision whether to merge the code based on Test_result.

Maybe this is equivalent to the `push:` trigger but I am not sure.

chrishkchris:
Sometime don't know why only
`Native-Ubuntu / build-cpptest-on-cpu (pull_request)`  is here above
But 
`Native-Ubuntu / build-cpptest-on-cpu (push)` is not here (e.g. in this PR)

chrishkchris:
I think I got the different between "push" and "pull_request" clearer now:

From this webpage: https://frontside.com/blog/2020-05-26-github-actions-pull_request/
This page says that:
"This difference means that a pull_request workflow ref would look like refs/remotes/pull/##/merge whereas a push workflow would be refs/heads/branch_name. This explains why the SHA of a push workflow matches the commit that triggered the workflow, whereas the SHA of a pull_request workflow does not; instead **the SHA of the pull_request is the resulting commit that was created from merging the base to the head**."

nudles:
> 
> 
> This PR removes Travis CI since we are using Github Actions now. :octocat:
> 
>     * We will not run the cpp and python lint commands:
> 
> 
> tool/linting/py.sh
> tool/linting/cpp.sh
> 
> Instead we can just use the output of Github's lgtm. Is this fine with everyone?
> 
>     * We do not have a github workflow that publishes the conda packages. If this is required, we can add the [Conda Package Publish Action](https://github.com/marketplace/actions/conda-package-publish-action) at the end of the conda workflow. Assuming there is secrets.ANACONDA_PASSWORD and secrets.ANACONDA_USERNAME defined in the [github secrets](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets) of the user who runs this workflow.
> 

How can we set the secrets? There is no `setting` tab for this project..

>     * Currently the github workflows use the latest Ubuntu only. If it is required to support another older versions, we can add them to the workflow. Travis CI was checking Ubuntu 14.04, 16.04 and 18.04.

moazreyad:
> How can we set the secrets? There is no `setting` tab for this project..

There are two solutions:

1. The secrets can be stored in the forked repository of the project by the person who wants to publish to conda. Then the conda publishing workflow will run only from this forked repository because it is the only place where the secrets are defined. 

2. The better option is to ask the Apache INFRA to add the secrets to the project.  In this case we don't store username and password, but we store the conda upload token. You may open a ticket similar to [this one](https://issues.apache.org/jira/browse/INFRA-20602) and send the token and explicit instructions as to how to create the token to root@apache.org. The same also can be done for publishing to PyPi.

PR #795, opened 2020-09-15
Update the Github Workflow for uploading conda packages

nudles:
The anaconda upload token is stored as a secret on github.


moazreyad:
I think it will be better to create a new separate workflow for publishing conda packages. In the conda package workflow, the build should not use ENABLE_TEST unless it is building a dev package for developers not a normal package for users. This will make the built code faster because it can use compiler optimizations which are not available in test mode and smaller because it does not contain the debug information. 

The packaging workflow (for conda, pypi, debian, docker, etc) should not contain the code coverage steps because (1) the code is usually not built with test enabled. (2) there is no need to upload a new coverage report for the same python test cases unless there are special test cases for the the package process itself.

The packaging workflow may not run with each pull request or merge on the dev branch, but they can run periodically every week for example or only when merging with the master branch.

nudles:
> 
> 
> I think it will be better to create a new separate workflow for publishing conda packages. In the conda package workflow, the build should not use ENABLE_TEST unless it is building a dev package for developers not a normal package for users. This will make the built code faster because it can use compiler optimizations which are not available in test mode and smaller because it does not contain the debug information.

> 
> The packaging workflow (for conda, pypi, debian, docker, etc) should not contain the code coverage steps because (1) the code is usually not built with test enabled. (2) there is no need to upload a new coverage report for the same python test cases unless there are special test cases for the the package process itself.
> 
> The packaging workflow may not run with each pull request or merge on the dev branch, but they can run periodically every week for example or only when merging with the master branch.


Currently, 
1. CPP test is done by native-macos and native-ubuntu for every PR, for which ENABLE_TEST is on.
2. Python test is done by conda for every PR, for which ENABLE_TEST is off.
3. conda package uploading is disabled on apache/singa because we cannot set the secret key for the uploading token.

I am fine to disable code coverage for 2 and 3, i.e., conda.yaml.
How should we disable it?

moazreyad:
> conda package uploading is disabled on apache/singa because we cannot set the secret key for the uploading token

Why we cannot set the secret key? Can we [ask INFRA to add the key](https://github.com/apache/singa/pull/790#issuecomment-690927600) as discussed in #790?

PR #806, opened 2020-10-09
fix README missing install link

lijiansong:


lijiansong:
Something maybe wrong with the CI checking of singa?

PR #824, opened 2021-01-12
Add new cpu conda build dockerfile and Add new readme

chrishkchris:
This PR has two purposes:

(i) Add a cpu conda build dockerfile, this is useful when the computer does not have gpu but we still want to build cpu package

(ii) Add new readme file on how to generate new conda packages


chrishkchris:
Adding new dockerfile and update readme without changing any existing code

PR #847, opened 2021-05-19
New folder cifar distributed cnn

naili-xing:


lzjpaul:
need to repull singa dev again before new changes

PR #848, opened 2021-05-19
add cifar100 and xceptionnet

naili-xing:
create folders for data modules for cifar_distributed_cnn


lzjpaul:
need to repull newest singa dev to include merge pull request #845 on May 19th

PR #853, opened 2021-05-21
autograd for mnist in cifar_distributed_cnn

naili-xing:
autograd for mnist in cifar_distributed_cnn


lzjpaul:
for autograd/mnist_cnn.py, need to import class CNN and functions, i.e., def check_dataset_exist(), def load_dataset(), def read_label_file(), etc

PR #898, opened 2021-11-22
Create singa-easy folder

NLGithubWP:


nudles:
Are we going to add singa-easy as a submodule of singa?

lzjpaul:
> Are we going to add singa-easy as a submodule of singa?

This pull request is not merged yet, need to discuss to finalize how singa-easy should be integrated/put into singa.

lzjpaul:
Putting into example folder should be better

PR #929, opened 2022-02-25
Add explanation module

NLGithubWP:
Add explanation module


lzjpaul:
the README.md should be in the same hierarchy of the singa_easy folder

PR #937, opened 2022-03-02
replace SGD optimizer with Adam for bloodmnist

zlheui:


lzjpaul:
Updating SGD with Adam for more steady training

PR #939, opened 2022-03-03
Update start.sh for medical image application

NLGithubWP:


lzjpaul:
added contents for start.sh

PR #941, opened 2022-03-04
Add stop script for medical image analysis application

NLGithubWP:


lzjpaul:
This stop script matches the start script.

PR #952, opened 2022-04-25
 add Alex net model for largedataset_cnn

wannature:
add Alex net model for largedataset_cnn


lzjpaul:
please create a model folder and then put the alexnet.py inside it, thanks

PR #953, opened 2022-04-26
add Alex net model for largedataset_cnn

wannature:
add Alex net model for largedataset_cnn


lzjpaul:
Yes, this commit is correct.

The newly added alexnet model is in the model folder

PR #954, opened 2022-04-27
add xceptionnet implementation for largedataset_cnn example

zlheui:


lzjpaul:
This commit is based on the latest dev, which is correct

PR #956, opened 2022-05-06
add cnn implementations for large dataset example

wannature:
add cnn implementations for large dataset example


lzjpaul:
added all four models, i.e., alexnet, cnn, resnet and xceptionet for large dataset example

PR #995, opened 2022-08-30
Update inline comments for cifar_distributed_cnn resnet_cifar10

KimballCai:
Update inline comments for cifar_distributed_cnn resnet_cifar10


lzjpaul:
Please use the latest dev for new commits

PR #1016, opened 2022-12-02
Disable dataset shape checking for cnn example

zlheui:


lzjpaul:
CLose the PR to redo it

PR #1059, opened 2023-05-30
malaria with cnn model

liuchangshiye:
Example of malaria dataset and CNN model


lzjpaul:
please re-pull-request with readme file

PR #1060, opened 2023-05-30
readme for malaria prediction with cnn model

liuchangshiye:
Readme for malaria prediction with CNN model


lzjpaul:
merge "readme for malaria prediction with cnn model"

PR #1071, opened 2023-07-15
add cnn model for malaria detection

liuchangshiye:
add cnn model for malaria detection


lzjpaul:
reopen pull request #1072 with updated dev-postgresql branch

PR #1098, opened 2023-09-16
Update the training file for dynamic models

Wu-Junran:
Optimize model training in multi-layer perceptron.


lzjpaul:
make the user account name consistent

PR #1119, opened 2023-11-01
Update python versions for wheel files

lzjpaul:
Update python versions for wheel files


tool/docker/devel/centos6/cuda10/Dockerfile.manylinux_2_28_x86_64_test_wheels
```diff
@@ -0,0 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
nudles, line 1: remove this file?

lzjpaul: ok, I will remove this one

PR #1142, opened 2024-03-08
Update image

NLGithubWP:


lzjpaul:
update the images for some components, e.g., mlcask, leads

PR #1146, opened 2024-03-09
Dev postgresql

NLGithubWP:
1. Remove the large wheel file.
2. Update the Dockerfile for both polarDB and PostgreSQL.
3. Update README.md
4. Change the name to'singa_trails_polardb'


lzjpaul:
Update the singa wheel files for polardb and postgresql

lzjpaul:
Fixed the bugs for Singa running on top of both polarDB and PostgreSQL

PR #1151, opened 2024-03-12
Init dynamic model slicing examples, add README.md

solopku:
Initialize dynamic model slicing examples.


solopku:
Add train mpi implementations.

PR #1206, opened 2024-08-17
update the running script in the readme file of the transformer example

Zrealshadow:


lzjpaul:
This feature was updated in https://github.com/apache/singa/pull/1206

PR #1248, opened 2024-12-24
Add the running script for TED CT Detection

lemonviv:
This pull request adds the running script for TED CT Detection.


lzjpaul:
Modify the train.py for proper model import

PR #1284, opened 2025-03-09
Add the dataset for the cardiovascular disease detection

serakiepiphany:
Add the dataset for the cardiovascular disease detection


lzjpaul:
Need to condense them into one commit

lzjpaul:
Need to condense them into one commit

PR #1287, opened 2025-03-15
Update the CMakeList file for V 5.0.0

lzjpaul:
Update the CMakeList file for V 5.0.0


nudles:
need to update https://github.com/apache/singa/blob/master/setup.py#L86 as well?

lzjpaul:
> need to update https://github.com/apache/singa/blob/master/setup.py#L86 as well?

Yes, also need to update the setup.py, I create a new pull request.

lzjpaul:
Create a new pull request to update the cmake and setup files

PR #1388, opened 2026-07-03
restructure the msmlp folder

LearnerLiyf:


examples/msmlp/native.py
```diff
@@ -1,137 +0,0 @@
-#
```
moazreyad, line 1: This file appears to be identical to:

https://github.com/apache/singa/blob/dev-postgresql/examples/mlp/native.py

So I think it should be safe to delete this duplicate file.

